PR #1, opened 2018-06-12
Multibyte sequence handling

NiLuJe:
Assuming input is valid UTF-8, do something useful with it instead of eating valuable column space with blanks ;).


NiLuJe:
This isn't perfect, but thanks to the hobbled libc setup on Kobos, we can't really do much better without basically reinventing the wheel (and it's a large & complex wheel, so, no thanks.).
See my numerous rants in the comments & commit messages for more details.

PR #2, opened 2018-06-13
Handle linefeeds

NiLuJe:
As in, the LF character (0x0A), not the escape code (\n).

### DIFF
--- a/fbink.c
+++ b/fbink.c
@@ -155,14 +155,6 @@ static char*
     font8x8_get_bitmap(uint32_t codepoint)
 {
 	// Get the bitmap for that ASCII character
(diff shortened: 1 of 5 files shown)


NiLuJe:
Meaning if you want to print your own stuff with linefeeds, get something else to handle the escade code conversion.

i.e., `fbink "$(echo -e "This\nis\nprinted\nover\nmultiple\nlines")" -x 1 -y 10`

NiLuJe:
The LF character gets rendered as a blank space, and I'm keeping it that way, for... reasons.

(It makes padding prettier, and the code simpler to deal with).

PR #3, opened 2018-06-15
Add some more fonts

NiLuJe:
So that my various experiments at least end up as being somewhat useful ;).


NiLuJe:
All the extra font in this set are actually variants of the fantastic Unscii (http://pelulamu.net/unscii) :).

PR #14, opened 2018-09-18
Kobo viewport

NiLuJe:
Handles the utter insanity that is devices with rows of pixels hidden by a bezel *by fucking design*.

That's the H2O, for sure (top 11 pixels, according to Nickel. Depending on the shitty fit of the IR array on your specific device, you might actually be able to see a few of those, if you're being ornery and looking at it at an extreme angle).
The Aura, quite likely, although I've never managed to get a proper screenshot to confirm the exact amount of pixels (most likely around 10 at the bottom). It makes even less sense than on the H2O, because unlike the H2O, I think the Aura's using a capacitive grid?!
Madness.

And the H2O² is a question mark... I'd have hoped they wouldn't have kept this bit of design insanity, but the fact is, we're using the same viewport as the H2O on those in KOReader, and someone has yet to complain about those 11 blank pixels on top, so... -_-".

In any case, right now, that's only enabled on the H2O, because in this instance, I'll only believe what I can see with my own eyes ;).

----

In a somewhat related fashion, piggyback on that to ensure that rows are "centered" on the usable screen height when we can't perfectly fit the final line on screen (which happens to now be pretty much always on the H2O, because 1429 is way less friendly to multiples of 8 than 1440 ;D).

This should ensure that row 0 is *always* usable (& visible), but it also means that stuff will very likely shift vertically by a few pixels, compared to the current behavior, where row 0, col 0 is always at (0, 0).

(It's a simple translation, it will *not* generate background padding like, well, padding does ;)).

That's enabled everywhere it's needed, but it will NEVER affect images. Only the "hardware" viewport is honored in the image codepath.

----

All this can optionally be toggled off, to restore the current behavior.


NiLuJe:
All this piggybacks on the h/voffset feature, which I'm hoping I've tested sufficiently to squish all the corner-cases...

This means that if anything feels or looks wonky after this merge, I'm interested ;).

PR #15, opened 2018-09-22
Add ability to print image from image data in memory

shermp:
This PR adds the ability for FBInk to print an image given a pointer to an image in memory.

What I've done is extract the STB image loading out of the main printing function. The main printing function has been renamed to `fbink_print_image_data()`. The STB loading function keeps the original `fbink_print_image()` function signature, to maintain backwards compatibility.

`fbink_print_image()` now loads the image from a file, and then calls `fbink_print_image_data()`. Alternatively, users can call `fbink_print_image_data()` directly if they do not wish to load an image from file.

`fbink_get_image_channels()` has been added to assist callers of `fbink_print_image_data()`, so that they know what image format should be supplied.

Tested on my H2O, with RGBA at least. File loading still appears to work fine, so hopefully I haven't broken anything!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "cstdlib": "c"
+    }
+}
--- a/fbink.c
+++ b/fbink.c
@@ -2977,7 +2977,7 @@ int
 // Draw an image on screen
 int
     fbink_print_image_data(int fbfd    UNUSED_BY_MINIMAL,
-		      const char* data UNUSED_BY_MINIMAL,
+		      unsigned char* data UNUSED_BY_MINIMAL,
 		      short int x_off UNUSED_BY_MINIMAL,
 		      short int y_off    UNUSED_BY_MINIMAL,
 			  int w UNUSED_BY_MINIMAL,
--- a/fbink.c
+++ b/fbink.c
@@ -2826,6 +2826,36 @@ cleanup:
 }
 
 int
+	fbink_get_image_channels(const FBInkConfig* fbink_config UNUSED_BY_MINIMAL)
+{
+#ifdef FBINK_WITH_IMAGE
+	int req_n;
+	switch (vInfo.bits_per_pixel) {
+		case 4U:
+			req_n = 1 + !fbink_config->ignore_alpha;
+			break;
+		case 8U:
+			req_n = 1 + !fbink_config->ignore_alpha;
+			break;
+		case 16U:
+			req_n = 3 + !fbink_config->ignore_alpha;
+			break;
+		case 24U:
+			req_n = 3 + !fbink_config->ignore_alpha;
+			break;
+		case 32U:
+		default:
+			req_n = 3 + !fbink_config->ignore_alpha;
+			break;
+	}
+	return req_n;
+#else
+	fprintf(stderr, "[FBInk] Image support is disabled in this FBInk build!\n");
+	return ERRCODE(ENOSYS);
+#endif    // FBINK_WITH_IMAGE
+}
+
+int
     fbink_print_image(int fbfd    UNUSED_BY_MINIMAL,
 		      const char* filename UNUSED_BY_MINIMAL,
 		      short int x_off UNUSED_BY_MINIMAL,
@@ -2939,8 +2969,10 @@ int
 	LOG("Requested %d color channels, image had %d.", req_n, n);
 
 	return fbink_print_image_data(fbfd, data, x_off, y_off, w, h, n, req_n, fbink_config);
-
-#endif
+#else
+	fprintf(stderr, "[FBInk] Image support is disabled in this FBInk build!\n");
+	return ERRCODE(ENOSYS);
+#endif    // FBINK_WITH_IMAGE
 }
 // Draw an image on screen
 int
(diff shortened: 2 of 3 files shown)


NiLuJe:
I'll take a deeper look at it tomorrow, so just a quick & random set of comments:

That's roughly how I'd imagined doing it (not that there are billions of way to go about it anyway ;)), although I hadn't thought to bother with the req_n bit, possibly thinking: feed me rgba or gtfo ;).

The fact that data is freed by the function and not the caller might feel non-inintuitive for `fbink_print_image_data()` callers  though, but OTOH I can't think of a nice way to make the distinction...

shermp:
As far as the RGBA matter, I was trying to keep things as close as possible to how they already were, FBInk currently expects stbi_load and friends to convert images to the correct format.

I'm embarrassed to say I didn't think of where the data was freed (oops, I've been in go land too much the past several days :p ). However, the simple solution would probably be to pull the freeing out of `fbink_print_image_data()`, and let the caller (and `fbink_print_image()`) free the data. I'll see how easy this is, and update the PR accordingly

NiLuJe:
Yep, that works ;).

shermp:
From what I can see in the code, req_n is used a lot. n is the value that is basically unused. From reading the stbi docs, req_n determines the overall size in bytes of the data buffer, so it's pretty important. From what I understand, the value stored in the pointer n is purely informational to say "this is the format the original image was in"

NiLuJe:
AFAICT, it still expects that data to be encoded in one of the supported formats, and not raw scanlines, so it doesn't need to know about the channel count to decode the data properly: the headers of the image format will answer that.

We indeed do care about req_n more than n, but it's interesting for stbi itself, to know how it'll have to *output* the image, and to FBInk, because that in turn dictates how we'll process it.

But the caller doesn't need to care about any of that: it just fills a buffer with the full data of an image encoded in a supported format, but the details of that can stay opaque, the only thing stbi needs to know is how much data there is (i.e., the buffer's length).

NiLuJe:
Because if we just have raw scanlines as input, we don't need stbi at all: just tell FBInk what you're feeding it, and one of many branches in that giant function will do the right thing :D.

shermp:
Note that I didn't go through the whole FBInk image print monstrosity myself, are you saying that it's happy to accept RGBA scanline despite whatever format is required for the framebuffer?

Because yes, the goal is to feed raw scanlines to FBInk. My usage is to draw directly to an image.RGBA in Go, and then feed the raw pixels directly into FBInk. No need for intermediary conversion steps then :)

I'll let you sleep on it, and you can give me your thought's when you have a chance to have a proper look.

I'm surprised you're awake TBH! It's late afternoon here on the other side of the world.

NiLuJe:
Yeah, that's what I finally got (I was initially under the mistaken impression you wanted to use stbi_load_from_memory(), which might explain the crossed wires :D).

So, in that case, yeah: just tell FBInk the width, height, and how many components there are, and things will mostly be okay.

With a somewhat large caveat right now: 8bpp fbs will only know how to deal with n 1 or 2; and >= 16bpp n 3 or 4.

(And *now* I get why you wanted to know req_n :D).

shermp:
All good.

In many ways `stbi_load_from_memory()` would certainly be easier, but hey, the less encode/decode steps, the better IMO.

NiLuJe:
Definitely ;).

So that leaves one question: do we require G8/G8A input on 4/8bpp & RGB/RGBA input on 16/24/32, which a simple function outputting what the device expects, or do we go the extra mile and handle any possible input? (It'll make that giant mess of a function even messier, and it's mildly annoying, because we have to handle switching to/from grayscale ourselves).

shermp:
Probably the former would be my thought. Are there even many (any?) modern ereaders that are 4/8bpp anymore? Note, I am a biased Kobo owner, so I don't know what those Kindle heathens are up to :D

NiLuJe:
Given that all Kindles fall under the 4/8 bpp case, and all Kobo fall under the 16/32 one, I can live with the first one (it's also way less boring work for me, so, that's a plus ;D).

NiLuJe:
Then we're agreed ;).

So to recap:

* Keep fbink_get_image_channels, but simplify it to basically return two different values: grayscale or not. I'd maybe rename it to something slightly more explicit?

* fbink_print_image_data still needs to know the exact channel count, and error out if g8 is passed to a Kobo (!grayscale), or rgb to a Kindle (grayscale).

shermp:
Sounds good.

Maybe a simple boolean function `fbink_greyscale_image_required()` or similar?

NiLuJe:
Yup, that's what jumped to mind after having written that ;).

NiLuJe:
Note to self: instead of trying to handle all n/req_n combos in the rendering stage, we could simply fudge the data buffer if a mismatch is detected (i.e., do the grayscale tweak in a new buffer, and render that one).

NiLuJe:
@shermp: Now that, despite my addled brain, we've agreed on what to do, do you want me to finish polishing this, or do you have some unpublished code?

(Sorry about the delay, I've been a bit busy these last few weeks ;)).

shermp:
I haven't done anything with this since, so you can polish it up if you wish.

I've been working on my other project in the meantime, and had a week where IRL issues got in the way of doing anything.

NiLuJe:
No worries, just wanted to make sure no work/time would be wasted ;).

NiLuJe:
Implemented in #18 ;).

Thanks again for all your input, it's very much appreciated!

PR #20, opened 2018-10-26
[WIP] OpenType Support

shermp:
Hi NiLuJe

I decided to see if I could figure out this whole OpenType/TrueType thing...

I'm opening this pull request relatively early, so I hopefully don't go too far down the wrong track if I've done stuff you don't agree with and/or want changed. Consider this Not Done Yet.

With that out the way, here's what's working:
* Printing strings to screen in arbitrary font sizes (I'm using points)
* Line breaking using the libunibreaks library, which implements the Unicode UAX 14 algorithm. Linebreaking honors hard line breaks (line feeds etc).
* Users can specify arbitrary margins to set the maximum printable area. Currently using percentages, although I'm mulling the idea of moving to pixels directly.
* ~~Centered text~~ Oops, this seems a bit broken if I have linebreaks in there.

What's not implemented:
* Overlays, foreground/background colors
* Vertical centering
* Probably others.

I've done away with the rows/columns concept, as it doesn't make much sense when using proportional fonts. I've also currently created a new struct for the OT settings. We may wish to integrate these to the main FBInkConfig struct.

Nothing has been optimized yet. We may want to look at caching glyph bitmaps for example (although, that will increase memory usage).

I also haven't handled any rotation shenanigans, I'm not sure if what I've done requires this.

I look forward to getting your feedback on this. No rush though, you are allowed to sleep!

(Also, if I may make an observation on the current printing code? The rows/columns concept seems to be more trouble than its worth IMHO)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "stb"]
 	path = stb
 	url = https://github.com/nothings/stb.git
+[submodule "libunibreak"]
+	path = libunibreak
+	url = https://github.com/adah1972/libunibreak
--- a/CREDITS
+++ b/CREDITS
@@ -8,10 +8,13 @@ This project makes use of a number of third-party libraries, which we'll acknowl
 		Released into the public domain.
 		(https://www.cprogramming.com/tutorial/unicode.html)
 
-	* stb_image (Image loader), by Sean Barrett,
+	* stb_image (Image loader) and stb_truetype (font renderer), by Sean Barrett,
 		Released into the public domain.
 		(http://nothings.org/stb)
 
+	* libunibreak (Unicode line breaking implementation) by Wu Yongwei, et al
+		Licensed under the Zlib license
+		(https://github.com/adah1972/libunibreak)
 
 As well as a number of third-party resources (namely, font files):
 
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ else
 	RANLIB?=gcc-ranlib
 endif
 
-DEBUG_CFLAGS=-O0 -fno-omit-frame-pointer -pipe -g
+DEBUG_CFLAGS=-Og -fno-omit-frame-pointer -pipe -g
 # Fallback CFLAGS, we honor the env first and foremost!
 OPT_CFLAGS=-O2 -fomit-frame-pointer -pipe
 
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ else
 	RANLIB?=gcc-ranlib
 endif
 
-DEBUG_CFLAGS=-Og -fno-omit-frame-pointer -pipe -g
+DEBUG_CFLAGS=-O0 -fno-omit-frame-pointer -pipe -g
 # Fallback CFLAGS, we honor the env first and foremost!
 OPT_CFLAGS=-O2 -fomit-frame-pointer -pipe
 
@@ -179,10 +179,16 @@ ifdef MATHS
 	EXTRA_CPPFLAGS+=-DFBINK_WITH_MATHS
 	LIBS+=-lm
 endif
-
+ifndef MINIMAL
+	LIBS+=-lm
+endif
 ##
 # Now that we're done fiddling with flags, let's build stuff!
-LIB_SRCS=fbink.c utf8/utf8.c
+ifdef MINIMAL
+	LIB_SRCS=fbink.c utf8/utf8.c
+else
+	LIB_SRCS=fbink.c utf8/utf8.c libunibreak/src/linebreak.c libunibreak/src/linebreakdata.c libunibreak/src/unibreakdef.c libunibreak/src/linebreakdef.c
+endif
 CMD_SRCS=fbink_cmd.c
 BTN_SRCS=button_scan_cmd.c
 # Unless we're asking for a minimal build, include the Unscii fonts, too
@@ -191,6 +197,7 @@ ifdef MINIMAL
 else
 	EXTRA_CPPFLAGS+=-DFBINK_WITH_FONTS
 	EXTRA_CPPFLAGS+=-DFBINK_WITH_IMAGE
+	EXTRA_CPPFLAGS+=-DFBINK_WITH_OPENTYPE
 	# Connect button scanning is Kobo specific
 	ifndef KINDLE
 		ifndef CERVANTES
@@ -235,7 +242,7 @@ $(OUT_DIR)/%.o: %.c
 	$(CC) $(CPPFLAGS) $(EXTRA_CPPFLAGS) $(CFLAGS) $(EXTRA_CFLAGS) $(SHARED_CFLAGS) -o $@ -c $<
 
 outdir:
-	mkdir -p $(OUT_DIR)/shared/utf8 $(OUT_DIR)/static/utf8
+	mkdir -p $(OUT_DIR)/shared/utf8 $(OUT_DIR)/static/utf8 $(OUT_DIR)/shared/libunibreak/src $(OUT_DIR)/static/libunibreak/src
 
 all: outdir static
 
@@ -348,8 +355,10 @@ clean:
 	rm -rf Release/*.so*
 	rm -rf Release/shared/*.o
 	rm -rf Release/shared/utf8/*.o
+	rm -rf Release/shared/libunibreak/src/*.o
 	rm -rf Release/static/*.o
 	rm -rf Release/static/utf8/*.o
+	rm -rf Release/static/libunibreak/src/*.o
 	rm -rf Release/*.o
 	rm -rf Release/fbink
 	rm -rf Release/button_scan
--- a/Makefile
+++ b/Makefile
@@ -366,8 +366,10 @@ clean:
 	rm -rf Debug/*.so*
 	rm -rf Debug/shared/*.o
 	rm -rf Debug/shared/utf8/*.o
+	rm -rf Debug/shared/libunibreak/src/*.o
 	rm -rf Debug/static/*.o
 	rm -rf Debug/static/utf8/*.o
+	rm -rf Debug/static/libunibreak/src/*.o
 	rm -rf Debug/*.o
 	rm -rf Debug/fbink
 	rm -rf Debug/button_scan
--- a/fbink.c
+++ b/fbink.c
@@ -62,6 +62,7 @@
 #endif
 
 #ifdef FBINK_WITH_OPENTYPE
+#	include <sys/stat.h>
 // stb_truetype needs maths, and so do we to round to the nearest pixel
 #	include <math.h>
 #	define STB_TRUETYPE_IMPLEMENTATION
@@ -2059,17 +2060,17 @@ int
 	FILE *f = fopen(fp, "rb");
 	unsigned char *data = NULL;
 	if (f) {
-		fseek(f, 0, SEEK_END);
-		long size = ftell(f);
-		fseek(f, 0, SEEK_SET);
-		data = calloc(1, size);
+		int fd = fileno(f);
+		struct stat st;
+		fstat(fd, &st);
+		data = calloc(1, st.st_size);
 		if (!data) {
 			fclose(f);
 			otInit = false;
 			return(ERRCODE(EXIT_FAILURE));
 		}
-		size_t read = fread(data, 1, size, f);
-		if (read != size) {
+		size_t read = fread(data, 1, st.st_size, f);
+		if (read != st.st_size) {
 			free(data);
 			fclose(f);
 			otInit = false;
@@ -2785,6 +2786,8 @@ int
 		//Note, we do a lot of casting floats to ints, so silence those GCC warnings
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wfloat-conversion"
+# 	pragma GCC diagnostic ignored "-Wconversion"
+#	pragma GCC diagnostic ignored "-Wbad-function-cast"
 
 	// Has fbink_init_ot() been called yet?
 	if (!otInit) {
@@ -2804,7 +2807,7 @@ int
 	// Declare buffers early to make cleanup easier
 	FBInkOTLine* lines = NULL;
 	char * brk_buff = NULL;
-	char* fmt_buff = NULL;
+	unsigned char* fmt_buff = NULL;
 	unsigned char* line_buff = NULL;
 	unsigned char* glyph_buff = NULL;
 
@@ -2842,12 +2845,15 @@ int
 		size_pt = 12;
 	}
 	// TODO: handle ppi properly
-	int ppi = 265;
+	unsigned int ppi = 265;
 	// Given the ppi, convert point height to pixels. Note, 1pt is 1/72th of an inch
-	int font_size_px = (int)(ppi / 72.0f * size_pt);
+	unsigned int font_size_px = (unsigned int)(ppi / 72.0f * size_pt);
 
+	// This is a pointer to whichever font is currently active. It gets updated for every
+	// character in the loop, as needed.
 	stbtt_fontinfo *curr_font = NULL;
 
+	// A struct to hold some metrics for each font as a whole
 	struct font_v_metrics {
 		float sf;
 		int asc;
@@ -2856,6 +2862,8 @@ int
 		int baseline;
 	};
 	int max_height = 0;
+	// Calculate some metrics for every font we have loaded.
+	// Please forgive the repetition here.
 	struct font_v_metrics rgMetrics, itMetrics, bdMetrics, bditMetrics;
 	if (otFonts.otRegular) {
 		rgMetrics.sf = stbtt_ScaleForPixelHeight(otFonts.otRegular, (float)font_size_px);
@@ -2893,16 +2901,16 @@ int
 			max_height = height;
 		}
 	}
-	// Get the Scale Factor used for most of the stbtt functions	
-	int num_lines = (int)(viewHeight / (uint32_t)max_height);
-	int baseline;
+	// Calculate the maximum number of lines we may have to deal with
+	unsigned int num_lines = (unsigned int)(viewHeight / (uint32_t)max_height);
+	
 	// And allocate the memory for it...
 	lines = calloc(num_lines, sizeof(FBInkOTLine));
 
 	// Now, lets use libunibreak to find the possible break opportunities in our string.
 
 	// Note: we only care about the byte length here
-	int str_len_bytes = strlen(string);
+	size_t str_len_bytes = strlen(string);
 	brk_buff = calloc(str_len_bytes + 1, sizeof(char));
 
 	init_linebreak();
@@ -2916,15 +2924,18 @@ int
 	}
 	// Lets find our lines! Nothing fancy, just a simple first fit algorithm, but we do
 	// our best not to break inside a word.
-	float sf;
-	int asc, dec, lg;
+
+	// Initialising the following three variables to make GCC happy.
+	float sf = 0.0;
+	int baseline = 0;
+	int lg = 0;
 	unsigned int chars_in_str = u8_strlen(string);
 	unsigned int c_index = 0;
 	unsigned int tmp_c_index = c_index;
 	uint32_t c;
 	unsigned short max_lw = area.br.x - area.tl.x;
 	printf("Max LW: %d\n", max_lw);
-	int line = 0;
+	unsigned int line = 0;
 	// adv = advance: the horizontal distance along the baseline to the origin of
 	//                the next glyph
 	// lsb = left side bearing: The horizontal distance from the origin point to
@@ -2938,7 +2949,7 @@ int
 		while (c_index < chars_in_str) {
 			// Check if we need to skip formatting characters
 			if (fmt_buff[c_index] == CH_IGNORE) {
-				c_index++;
+				u8_inc(string, &c_index);
 				continue;
 			} else {
 				switch (fmt_buff[c_index]) {
@@ -2964,9 +2975,9 @@ int
 				rv = ERRCODE(ENOENT);
 				goto cleanup;
 			}
-			// First, we check for a mandatory break
+			// We check for a mandatory break
 			if (brk_buff[c_index] == LINEBREAK_MUSTBREAK) {
-				unsigned char last_index = c_index;
+				unsigned int last_index = c_index;
 				// We don't want to print the break character
 				u8_dec(string, &last_index);
 				lines[line].endCharIndex = last_index;
@@ -3040,7 +3051,8 @@ int
 	FBInkCoordinates curr_point = { 0, 0 };
 	FBInkCoordinates ins_point = {0, 0};
 	FBInkCoordinates paint_point = {area.tl.x, area.tl.y};
-	int x0, y0, x1, y1, gw, gh, lw;
+	int x0, y0, x1, y1, gw, gh;
+	unsigned int lw;
 	uint32_t tmp_c;
 	unsigned char *lnPtr, *glPtr = NULL;
 	unsigned short start_x, start_y;
@@ -3050,10 +3062,10 @@ int
 	for (line = 0; lines[line].line_used; line++) {
 		printf("Line # %d\n", line);
 		lw = 0;
-		int ci;
+		unsigned int ci;
 		for (ci = lines[line].startCharIndex; ci <= lines[line].endCharIndex;) {
 			if (fmt_buff[ci] == CH_IGNORE) {
-				ci++;
+				u8_inc(string, &ci);
 				continue;
 			} else {
 				switch (fmt_buff[ci]) {
@@ -3123,7 +3135,7 @@ int
 			}
 			curr_point.x += (int)roundf(sf * adv);
 			if (ci < lines[line].endCharIndex) {
-				int tmp_i = ci;
+				unsigned int tmp_i = ci;
 				tmp_c = u8_nextchar(string, &tmp_i);
 				curr_point.x += (int)roundf(sf * stbtt_GetCodepointKernAdvance(curr_font, c, tmp_c));
 			}
--- a/fbink.c
+++ b/fbink.c
@@ -61,6 +61,27 @@
 #	pragma GCC diagnostic pop
 #endif
 
+#ifdef FBINK_WITH_OPENTYPE
+// stb_truetype needs maths, and so do we to round to the nearest pixel
+#	include <math.h>
+#	define STB_TRUETYPE_IMPLEMENTATION
+// stb_truetype is.... noisy
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#	pragma clang diagnostic ignored "-Wunknown-warning-option"
+#	pragma GCC diagnostic ignored "-Wcast-qual"
+#	pragma GCC diagnostic ignored "-Wcast-align"
+#	pragma GCC diagnostic ignored "-Wconversion"
+#	pragma GCC diagnostic ignored "-Wsign-conversion"
+#	pragma GCC diagnostic ignored "-Wduplicated-branches"
+#	pragma GCC diagnostic ignored "-Wunused-parameter"
+#	pragma GCC diagnostic ignored "-Wunused-function"
+#	pragma GCC diagnostic ignored "-Wbad-function-cast"
+#	include "stb/stb_truetype.h"
+#	pragma GCC diagnostic pop
+#	include "libunibreak/src/linebreak.h"
+#endif
+
 // Return the library version as devised at library compile-time
 const char*
     fbink_version(void)
@@ -2028,6 +2049,23 @@ int
 	// Don't skip any ioctls on a first init ;)
 	return initialize_fbink(fbfd, fbink_config, false);
 }
+// Load font from a provided buffer.
+// NOTE: the caller MUST NOT free the font_data buffer untill they know they will no longer need to print.
+//		 Should we create our own copy of the font data instead? Perhaps from a FILE pointer? Or filepath?
+int
+	fbink_init_ot(const unsigned char* font_data)
+{
+#ifdef FBINK_WITH_OPENTYPE
+	int rv = EXIT_SUCCESS;
+	otFontInfo = (const stbtt_fontinfo){ 0 };
+	if (!stbtt_InitFont(&otFontInfo, font_data, 0)) {
+		rv = ERRCODE(EXIT_FAILURE);
+	}
+	return rv;
+#else
+	return ERRCODE(ENOTSUP);
+#endif
+}
 
 // Dump a few of our internal state variables to stdout, for shell script consumption
 void
@@ -2593,6 +2631,271 @@ int
 	// Cleanup
 	free(buffer);
 	return rv;
+}
+
+int
+	fbink_print_ot(int fbfd, char* string, FBInkOTConfig* cfg)
+{
+#ifdef FBINK_WITH_OPENTYPE
+		//Note, we do a lot of casting floats to ints, so silence those GCC warnings
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wfloat-conversion"
+	// If we open a fd now, we'll only keep it open for this single print call!
+	// NOTE: We *expect* to be initialized at this point, though, but that's on the caller's hands!
+	bool keep_fd = true;
+	if (open_fb_fd(&fbfd, &keep_fd) != EXIT_SUCCESS) {
+		return ERRCODE(EXIT_FAILURE);
+	}
+
+	// Assume success, until shit happens ;)
+	int rv = EXIT_SUCCESS;
+
+	// Declare buffers early to make cleanup easier
+	FBInkOTLine* lines = NULL;
+	char * brk_buff = NULL;
+	unsigned char* line_buff = NULL;
+	unsigned char* glyph_buff = NULL;
+
+	// map fb to user mem
+	// NOTE: If we're keeping the fb's fd open, keep this mmap around, too.
+	if (!isFbMapped) {
+		if (memmap_fb(fbfd) != EXIT_SUCCESS) {
+			rv = ERRCODE(EXIT_FAILURE);
+			goto cleanup;
+		}
+	}
+	LOG("Printing OpenType text.");
+	// Sanity check the provided margins and calculate the printable area.
+	// We'll cap the margin at 90% for each side. Margins for opposing edges
+	// should sum to less than 100%
+	if (cfg->margins.top > 90 || cfg->margins.bottom > 90 || cfg->margins.left > 90 || cfg->margins.right > 90) {
+		rv = ERRCODE(ERANGE);
+		goto cleanup;
+	}
+	if (cfg->margins.top + cfg->margins.bottom >= 100 || cfg->margins.left + cfg->margins.right >= 100) {
+		rv = ERRCODE(ERANGE);
+		goto cleanup;
+	}
+	struct {
+		FBInkCoordinates tl;
+		FBInkCoordinates br;
+	} area = { 0 };
+	area.tl.x = (unsigned short)(cfg->margins.left / 100.0f * viewWidth);
+	area.tl.y = (unsigned short)(cfg->margins.top / 100.0f * viewHeight);
+	area.br.x = (unsigned short)(viewWidth - (uint32_t)(cfg->margins.right / 100.0f * viewWidth));
+	area.br.y = (unsigned short)(viewHeight - (uint32_t)(cfg->margins.bottom / 100.0f * viewHeight));
+
+	// TODO: handle ppi properly
+	int ppi = 265;
+	// Given the ppi, convert point height to pixels. Note, 1pt is 1/72th of an inch
+	int font_size_px = (int)(ppi / 72.0f * cfg->size_pt);
+	// Get the Scale Factor used for most of the stbtt functions
+	float sf = stbtt_ScaleForPixelHeight(&otFontInfo, (float)font_size_px);
+	// Get the max possible glyph height, and therefore calc the maximum number of lines
+	// that may be printed in the given dimensions.
+	int asc, dec, lg;
+	stbtt_GetFontVMetrics(&otFontInfo, &asc, &dec, &lg);
+	int num_lines = (int)(viewHeight / (uint32_t)roundf(sf * (asc - dec)));
+	int baseline = (int)roundf(sf * asc);
+	// And allocate the memory for it...
+	lines = calloc(num_lines, sizeof(FBInkOTLine));
+
+	// Now, lets use libunibreak to find the possible break opportunities in our string.
+
+	// Note: we only care about the byte length here
+	int str_len_bytes = strlen(string);
+	brk_buff = calloc(str_len_bytes + 1, sizeof(char));
+
+	init_linebreak();
+	set_linebreaks_utf8((utf8_t*)string, str_len_bytes + 1, "en", brk_buff);
+	LOG("Found linebreaks!");
+	// Lets find our lines! Nothing fancy, just a simple first fit algorithm, but we do
+	// our best not to break inside a word.
+	unsigned int chars_in_str = u8_strlen(string);
+	unsigned int c_index = 0;
+	unsigned int tmp_c_index = c_index;
+	uint32_t c;
+	unsigned short max_lw = area.br.x - area.tl.x;
+	printf("Max LW: %d\n", max_lw);
+	int line = 0;
+	// adv = advance: the horizontal distance along the baseline to the origin of
+	//                the next glyph
+	// lsb = left side bearing: The horizontal distance from the origin point to
+	//                          left edge of the glyph
+	int adv, lsb, curr_x;
+	for (line = 0; line < num_lines; line++) {
+		// Every line has a start character index and an end char index.
+		curr_x = 0;
+		lines[line].startCharIndex = c_index;
+		lines[line].line_used = true;
+		while (c_index < chars_in_str) {
+			// First, we check for a mandatory break
+			if (brk_buff[c_index] == LINEBREAK_MUSTBREAK) {
+				unsigned char last_index;
+				// We don't want to print the break character
+				u8_dec(string, &last_index);
+				lines[line].endCharIndex = last_index;
+				// We want our next line to start after this breakpoint
+				u8_inc(string, &c_index);
+				// And we're done processing this line
+				break;
+			}
+			c = u8_nextchar(string, &c_index);
+			// Note, these metrics are unscaled, we need to use our previously
+			// obtained scale factor (sf) to get the metrics as pixels
+			stbtt_GetCodepointHMetrics(&otFontInfo, (int)c, &adv, &lsb);
+			// If starting a line with a character that has a negative lsb
+			// eg 'j', move our start point a little.
+			if (curr_x == 0 && lsb < 0) {
+				curr_x += (int)roundf(sf * abs(lsb));
+			}
+			int x0, x1;
+			stbtt_GetCodepointBox(&otFontInfo, c, &x0, 0, &x1, 0);
+			int gw_from_origin = (int)roundf(sf * (x1 - x0)) - (int)roundf(sf * lsb);
+			printf("GW: %d\n", ((int)roundf(sf * (x1 - x0))));
+			printf("Calced LW: %d\n", (curr_x + gw_from_origin));
+			// Oops, we appear to have advanced too far :)
+			// Better backtrack to see if we can find a suitable break opportunity
+			if (curr_x + gw_from_origin > max_lw) {
+				// Set our final line character to the previous char in case we cannot
+				// find a suitable breakpoint
+				u8_dec(string, &c_index);
+				// Note, we need to do this a second time, to get the previous character, as
+				// u8_nextchar() 'consumes' a character.
+				u8_dec(string, &c_index);
+				lines[line].endCharIndex = c_index;
+				for (c_index; c_index > lines[line].startCharIndex; u8_dec(string, &c_index)) {
+					if (brk_buff[c_index] == LINEBREAK_ALLOWBREAK) {
+						lines[line].endCharIndex = c_index;
+						break;
+					}
+				}
+				u8_inc(string, &c_index);
+				// We are done with this line.
+				break;
+			}
+			curr_x += (int)roundf(sf * adv);
+			// Adjust our x position for kerning, because we can :)
+			if (string[c_index + 1]) {
+				tmp_c_index = c_index;
+				uint32_t c2 = u8_nextchar(string, &tmp_c_index);
+				curr_x += (int)roundf(sf * stbtt_GetCodepointKernAdvance(&otFontInfo, c, c2));
+			}
+		}
+		// We've run out of string! This is our last line.
+		if (c_index >= chars_in_str) {
+			u8_dec(string, &c_index);
+			lines[line].endCharIndex = c_index;
+			break;
+		}
+	}
+	LOG("Lines Found!");
+	// Hopefully, we have some lines to render!
+
+	// Create a bitmap buffer to render a single line. We don't render the glyphs directly to the
+	// fb here, as we need to do some simple blending, and it makes it easier to calculate our
+	// centering if required.
+	line_buff = calloc(max_lw * font_size_px, sizeof(char));
+	// We also don't want to be creating a new buffer for every glyph
+	glyph_buff = calloc(font_size_px * font_size_px * 2, sizeof(char));
+	if (!line_buff || !glyph_buff) {
+		goto cleanup;
+	}
+	FBInkCoordinates curr_point = { 0, baseline };
+	FBInkCoordinates ins_point = {0, baseline};
+	FBInkCoordinates paint_point = {area.tl.x, area.tl.y};
+	int x0, y0, x1, y1, gw, gh, lw;
+	uint32_t tmp_c;
+	for (line = 0; lines[line].line_used; line++) {
+		lw = 0;
+		int ci;
+		for (ci = lines[line].startCharIndex; ci <= lines[line].endCharIndex;) {
+			c = u8_nextchar(string, &ci);
+			stbtt_GetCodepointHMetrics(&otFontInfo, c, &adv, &lsb);
+			if (ci == lines[line].startCharIndex && lsb < 0) {
+				curr_point.x += (int)roundf(sf * abs(lsb));
+			}
+			stbtt_GetCodepointBitmapBox(&otFontInfo, c, sf, sf, &x0, &y0, &x1, &y1);
+			gw = x1 - x0;
+			printf("BMGW: %d\n", gw);
+			gh = y1 - y0;
+			ins_point.x = curr_point.x + (int)roundf(lsb * sf);
+			ins_point.y += y0;
+			lw = ins_point.x + gw;
+			stbtt_MakeCodepointBitmap(&otFontInfo, glyph_buff, gw, gh, 0, sf, sf, c);
+			// paint our glyph into the line buffer
+			unsigned char* lnPtr = line_buff + ins_point.x + (max_lw * ins_point.y);
+			unsigned char* glPtr = glyph_buff;
+			for (int j = 0; j < gh; j++) {
+				for (int k = 0; k < gw; k++) {
+					// 0 value pixels are transparent
+					if (glPtr[k] > 0) {
+						lnPtr[k] = glPtr[k];
+					}
+				}
+				// And advance one scanline. Quick! Hide! Pointer arithmetic
+				glPtr += gw;
+				lnPtr += max_lw;
+			}
+			curr_point.x += (int)roundf(sf * adv);
+			if (ci < lines[line].endCharIndex) {
+				int tmp_i = ci;
+				tmp_c = u8_nextchar(string, &tmp_i);
+				curr_point.x += (int)roundf(sf * stbtt_GetCodepointKernAdvance(&otFontInfo, c, tmp_c));
+			}
+			ins_point.y = baseline;
+			printf("LW: %d\n", lw);
+		}
+		// Right, we've rendered a line to a bitmap, time to display it.
+		if (cfg->is_centered) {
+			paint_point.x += (max_lw - lw) / 2;
+		}
+		FBInkColor color = { 0 };
+		unsigned short start_x = paint_point.x;
+		unsigned short start_y = paint_point.y;
+		unsigned char* lnPtr = line_buff;
+		for (int j = 0; j < font_size_px; j++) {
+			for (int k = 0; k < lw; k++) {
+				color.r = color.b = color.g = OT_INVERT_PIXEL(lnPtr[k]);
+				put_pixel(&paint_point, &color);
+				paint_point.x++;
+			}
+			lnPtr += max_lw;
+			paint_point.x = start_x;
+			paint_point.y++;
+		}
+		paint_point.y += (int)(sf * lg);
+#ifndef LINUX
+		// Woohoo, it's in our framebuffer! Let's refresh the screen.
+		struct mxcfb_rect region = { 0 };
+		region.left = start_x;
+		region.top = start_y;
+		region.width = lw;
+		region.height = font_size_px;
+		refresh(fbfd, region, WAVEFORM_MODE_AUTO, false);
+#endif
+		LOG("Printed Line!");
+		// And clear our line buffer for next use. The glyph buffer shouldn't
+		// need clearing, as stbtt_MakeCodepointBitmap() should overwrite it.
+		memset(line_buff, 0, (max_lw * font_size_px * sizeof(char)));
+	}
+
+	cleanup:
+		free(lines);
+		free(brk_buff);
+		free(line_buff);
+		free(glyph_buff);
+		if (isFbMapped && !keep_fd) {
+			unmap_fb();
+		}
+		if (!keep_fd) {
+			close(fbfd);
+		}
+	return rv;
+#	pragma GCC diagnostic pop
+#else
+	return ERRCODE(ENOTSUP);
+#endif
 }
 
 // Small public wrapper around refresh(), without the caller having to depend on mxcfb headers
--- a/fbink.c
+++ b/fbink.c
@@ -2777,7 +2777,7 @@ void
 }
 
 int
-	fbink_print_ot(int fbfd, char* string, FBInkOTConfig* cfg)
+	fbink_print_ot(int fbfd, char* string, FBInkOTConfig* cfg, FBInkConfig* fbCfg)
 {
 #ifdef FBINK_WITH_OPENTYPE
 		//Note, we do a lot of casting floats to ints, so silence those GCC warnings
@@ -2796,6 +2796,12 @@ int
 	if (!otInit) {
 		ELOG("[FBInk] No fonts have been loaded");
 		return ERRCODE(ENODATA);
+	}
+
+	// Just in case we receive a NULL pointer to the cfg struct
+	if (!cfg) {
+		ELOG("[FBInk] FBInkOTConfig expected. Got NULL pointer instead");
+		return ERRCODE(EXIT_FAILURE);
 	}
 
 	// If we open a fd now, we'll only keep it open for this single print call!
@@ -3188,6 +3194,20 @@ int
 	}
 	printf("Actual print height is %u\n", curr_print_height);
 
+	// Let's get some rendering options from FBInkConfig
+	uint8_t valign = NONE, halign = NONE, fgcolor = FG_BLACK, bgcolor = BG_WHITE;
+	bool is_inverted = false, is_overlay = false, is_bgless = false, is_fgless = false;
+	if (fbCfg) {
+		valign = fbCfg->valign;
+		halign = fbCfg->halign;
+		fgcolor = fbCfg->fg_color;
+		bgcolor = fbCfg->bg_color;
+		is_inverted = fbCfg->is_inverted;
+		is_overlay = fbCfg->is_overlay;
+		is_bgless = fbCfg->is_bgless;
+		is_fgless = fbCfg->is_fgless;
+	}
+
 	// Hopefully, we have some lines to render!
 
 	// Create a bitmap buffer to render a single line. We don't render the glyphs directly to the
@@ -3201,10 +3221,17 @@ int
 		rv = ERRCODE(EXIT_FAILURE);
 		goto cleanup;
 	}
+
 	// Setup the variables needed to render
 	FBInkCoordinates curr_point = { 0, 0 };
 	FBInkCoordinates ins_point = {0, 0};
 	FBInkCoordinates paint_point = {area.tl.x, area.tl.y};
+	// Set the vertical positioning now
+	if (valign == CENTER) {
+		paint_point.y += (print_height - curr_print_height) / 2;
+	} else if (valign == EDGE) {
+		paint_point.y += print_height - curr_print_height;
+	}
 	uint32_t tmp_c;
 	unsigned char *lnPtr, *glPtr = NULL;
 	unsigned short start_x, start_y;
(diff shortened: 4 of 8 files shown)


mePoP:
[Yay!](https://www.mobileread.com/forums/showpost.php?p=3760548&postcount=86)

shermp:
Note, I decided to go the **stb_truetype** route rather than try and use freetype2. Means we don't have to try and link against freetype (or figure out how to statically compile it), and FBInk already uses other STB libraries, so it was a simple matter of `#include "stb/stb_truetype"`.

Quality seems acceptable, probably not as nice as freetype2 would be, but good enough.

Still have to think about how we handle font loading. Current implementation is for the developer to load the font into memory and pass it to FBInk. It might be nicer if FBInk could accept a font file path.

NiLuJe:
And as far a glyph caching is, I'm guessing it'll all depend on use-cases and actual performance.

Story time:

In the cell rendered, I originally was decoding the font data, then scaling it into a buffer, then rendering that buffer.
Folding the scaling into the rendering pass (and getting rid of a buffer & copy in the process) netted me a noticeable performance increase.

NiLuJe:
I kind of appreciate the OT struct being on its own, as FBInkConfig is already starting to look like a giant mess ;).
Again, no strong feelings on that one, though.

NiLuJe:
And I definitely get the STB choice, that'd probably have been my first choice since the image loading code, too ;).

(Although I am rather well versed in compiling FT :D).

NiLuJe:
And a final comment in the "not to forget when you're done" box: adding a libunibreak mention to the CREDITS file ;).

shermp:
Hey, thanks for the feedback!

My only concern with passing font paths instead of an already loaded buffer is that I will need to add a function like `fbink_close_ot() to free the allocated memory. On the other hand, it would remove any ambiguity.

I'm leaning towards changing the margins to pixels as well. That means it becomes meaningful to return the Y position of the next row of text that callers can set on the next print call.

Thanks for the heads up on the macro. I never was that fond of it, but my bitwise-fu isn't always the greatest!

As to glyph caching, I'm really not certain. Generating them is certainly more intensive than a simple integer scaling however, as there's floating point math and anti-aliasing going on. On the other hand, even with the current method, text is basically appearing on the screen instantly so...

shermp:
Hmm, maybe we need a "get viewinfo" function to return basic resolution info (without requiring getting the full state)? Because if I allow the user to specify margins as pixels, they probably would like to know how many pixels are available first!

NiLuJe:
Yep, having an open/close pair would probably make it more obvious ;).

I'm not opposed to a specific getter for the resolution and/or viewport, but how were you envisioning it: returning a new struct that only contains a subset of the full state, or setting uint pointers?
Or something else? ;).

shermp:
Either method could work I suppose. I've gotten used to setting pointers with the STB libraries.

NiLuJe:
Yeah, that sounds like the least intrusive way to handle that ;).

shermp:
So... um... erm.... I have a confession to make.

I kind of went and added font variant support (bold, italics, bold italics), as well as a very rudimentary "markdown style" parser.

Ummm... Oops? :D

fontvariants

I'm pushing what I've done so far. Fair warning though, there are Problems. Namely segfaults. But only with certain combinations and lengths of text. In other words, I probably need to return to school, because I'm obviously having trouble with my sums. :p 

If you spot any obvious errors, please let me know. My brain has turned to mush...

NiLuJe:
:D 

I'll try to take a look at it this afternoon, with my old friend gdb if need be ;).

NiLuJe:
Speaking of crashes, it also crashes with an empty string ;).

NiLuJe:
FWIW, I'm testing with a quick'n dirty piggyback in fbink_cmd:

```
diff --git a/fbink_cmd.c b/fbink_cmd.c
index fabf953..c2a0193 100644
--- a/fbink_cmd.c
+++ b/fbink_cmd.c
@@ -690,11 +690,23 @@ int
                                    fbink_config.fontname,
                                    fbink_config.fontmult);
                        }
+                       // TTF!
+                       FBInkOTConfig cfg = { 0 };
+                       //cfg.is_formatted = true;
+                       cfg.margins.top = 5;
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Bold.ttf", FNT_BOLD);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-BoldItalic.ttf", FNT_BOLD_ITALIC);
+                       fbink_print_ot(fbfd, string, &cfg);
+                       fbink_free_ot_fonts();
+                       /*
                        if ((linecount = fbink_print(fbfd, string, &fbink_config)) < 0) {
                                fprintf(stderr, "Failed to print that string!\n");
                                rv = ERRCODE(EXIT_FAILURE);
                                goto cleanup;
                        }
+                       */
                        // NOTE: Don't clobber previous entries if multiple strings were passed...
                        //       We make sure to trust print's return value,
                        //       because it knows how much space it already took up ;).
```

NiLuJe:
Haven't managed to break anything else after a few quick tests ;).

```diff
diff --git a/fbink.c b/fbink.c
index 36f32eb..7d563ea 100644
--- a/fbink.c
+++ b/fbink.c
@@ -2114,8 +2114,8 @@ int
 }
 
 // Free an individual OpenType font structure
-void* 
-       free_ot_font(stbtt_fontinfo* font_info) 
+void*
+       free_ot_font(stbtt_fontinfo* font_info)
 {
        if (font_info) {
                free(font_info->data); // This is the font data we loaded
@@ -2712,38 +2712,24 @@ int
 // This is **bold** text.
 // This is ***bold italic*** text.
 // As well as their underscore equivalents
-void 
+void
        parse_simple_md(char* string, int size, unsigned char* result)
 {
        int ci = 0;
+       char ch;
        bool is_italic = false;
        bool is_bold = false;
        while (ci < size) {
-               switch (string[ci]) {
+               printf("ci: %d (< %d) is %c\n", ci, size, string[ci]);
+               switch (ch = string[ci]) {
                case '*':
-                       if (ci + 1 < size && string[ci + 1] == '*') {
-                               if (ci + 2 < size && string[ci + 2] == '*') {
-                                       is_bold = !is_bold;
-                                       is_italic = !is_italic;
-                                       result[ci] = CH_IGNORE;
-                                       result[ci + 1] = CH_IGNORE;
-                                       result[ci + 2] = CH_IGNORE;
-                                       ci += 3;
-                                       break;
-                               }
-                               is_bold = !is_bold;
-                               result[ci] = CH_IGNORE;
-                               result[ci + 1] = CH_IGNORE;
-                               ci += 2;
-                               break;
-                       }
-                       is_italic = !is_italic;
-                       result[ci] = CH_IGNORE;
-                       ci++;
-                       break;
                case '_':
-                       if (ci + 1 < size && string[ci + 1] == '_') {
-                               if (ci + 2 < size && string[ci + 2] == '_') {
+                       // FIXME: _ handling is problematic, because it'll catch in-word underscores when it shouldn't,
+                       //          since there's no closing tag ;).
+                       if (ci + 1 < size && string[ci + 1] == ch) {
+                               printf("ci: %d && ci + 1 == %c\n", ci, ch);
+                               if (ci + 2 < size && string[ci + 2] == ch) {
+                                       printf("ci: %d && ci + 2 == %c\n", ci, ch);
                                        is_bold = !is_bold;
                                        is_italic = !is_italic;
                                        result[ci] = CH_IGNORE;
@@ -2902,7 +2888,7 @@ int
        }
        // Calculate the maximum number of lines we may have to deal with
        unsigned int num_lines = (unsigned int)(viewHeight / (uint32_t)max_height);
-
+
        // And allocate the memory for it...
        lines = calloc(num_lines, sizeof(FBInkOTLine));
 
@@ -2921,7 +2907,7 @@ int
        LOG("Found linebreaks!");
 
        // Parse our string for formatting, if requested
-       fmt_buff = calloc(str_len_bytes, sizeof(char));
+       fmt_buff = calloc(str_len_bytes + 1, sizeof(char));
        if (!fmt_buff) {
                rv = ERRCODE(EXIT_FAILURE);
                goto cleanup;
```

So, basically, just making sure the result buffer is NULL-terminated, and avoiding code duplication between */_ handling.
With an obvious FIXME for _ (which also applies to basically any kind of unbalanced tag, even if the underscore case is probably the only one that can legitimately happen outside of md snafus).

NiLuJe:
And with a pretty crappy workaround to the _ thing:

```diff
diff --git a/fbink.c b/fbink.c
index 36f32eb..1d2ecfc 100644
--- a/fbink.c
+++ b/fbink.c
@@ -2114,8 +2114,8 @@ int
 }
 
 // Free an individual OpenType font structure
-void* 
-       free_ot_font(stbtt_fontinfo* font_info) 
+void*
+       free_ot_font(stbtt_fontinfo* font_info)
 {
        if (font_info) {
                free(font_info->data); // This is the font data we loaded
@@ -2712,17 +2712,22 @@ int
 // This is **bold** text.
 // This is ***bold italic*** text.
 // As well as their underscore equivalents
-void 
+void
        parse_simple_md(char* string, int size, unsigned char* result)
 {
        int ci = 0;
+       char ch;
        bool is_italic = false;
        bool is_bold = false;
        while (ci < size) {
-               switch (string[ci]) {
+               printf("ci: %d (< %d) is %c\n", ci, size, string[ci]);
+               switch (ch = string[ci]) {
                case '*':
-                       if (ci + 1 < size && string[ci + 1] == '*') {
-                               if (ci + 2 < size && string[ci + 2] == '*') {
+               case '_':
+                       if (ci + 1 < size && string[ci + 1] == ch) {
+                               printf("ci: %d && ci + 1 == %c\n", ci, ch);
+                               if (ci + 2 < size && string[ci + 2] == ch) {
+                                       printf("ci: %d && ci + 2 == %c\n", ci, ch);
                                        is_bold = !is_bold;
                                        is_italic = !is_italic;
                                        result[ci] = CH_IGNORE;
@@ -2737,25 +2742,10 @@ void
                                ci += 2;
                                break;
                        }
-                       is_italic = !is_italic;
-                       result[ci] = CH_IGNORE;
-                       ci++;
-                       break;
-               case '_':
-                       if (ci + 1 < size && string[ci + 1] == '_') {
-                               if (ci + 2 < size && string[ci + 2] == '_') {
-                                       is_bold = !is_bold;
-                                       is_italic = !is_italic;
-                                       result[ci] = CH_IGNORE;
-                                       result[ci + 1] = CH_IGNORE;
-                                       result[ci + 2] = CH_IGNORE;
-                                       ci += 3;
-                                       break;
-                               }
-                               is_bold = !is_bold;
-                               result[ci] = CH_IGNORE;
-                               result[ci + 1] = CH_IGNORE;
-                               ci += 2;
+                       // Try to avoid flagging a single underscore in the middle of a word.
+                       if (ch == '_' && ci > 0 && string[ci - 1] != ' ' && string[ci + 1] != ' ' && string[ci - 1] != ch && string[ci + 1] != ch) {
+                               result[ci] = CH_REGULAR;
+                               ci++;
                                break;
                        }
                        is_italic = !is_italic;
@@ -2902,7 +2892,7 @@ int
        }
        // Calculate the maximum number of lines we may have to deal with
        unsigned int num_lines = (unsigned int)(viewHeight / (uint32_t)max_height);
-
+
        // And allocate the memory for it...
        lines = calloc(num_lines, sizeof(FBInkOTLine));
 
@@ -2921,7 +2911,7 @@ int
        LOG("Found linebreaks!");
 
        // Parse our string for formatting, if requested
-       fmt_buff = calloc(str_len_bytes, sizeof(char));
+       fmt_buff = calloc(str_len_bytes + 1, sizeof(char));
        if (!fmt_buff) {
                rv = ERRCODE(EXIT_FAILURE);
                goto cleanup;
```

NiLuJe:
Given stbtt's native bitmap format, we can get inversion pretty easily, so, here goes:

```diff
diff --git a/fbink.c b/fbink.c
index 36f32eb..6b2e4dc 100644
--- a/fbink.c
+++ b/fbink.c
@@ -2114,8 +2114,8 @@ int
 }
 
 // Free an individual OpenType font structure
-void* 
-       free_ot_font(stbtt_fontinfo* font_info) 
+void*
+       free_ot_font(stbtt_fontinfo* font_info)
 {
        if (font_info) {
                free(font_info->data); // This is the font data we loaded
@@ -2712,17 +2712,22 @@ int
 // This is **bold** text.
 // This is ***bold italic*** text.
 // As well as their underscore equivalents
-void 
+void
        parse_simple_md(char* string, int size, unsigned char* result)
 {
        int ci = 0;
+       char ch;
        bool is_italic = false;
        bool is_bold = false;
        while (ci < size) {
-               switch (string[ci]) {
+               printf("ci: %d (< %d) is %c\n", ci, size, string[ci]);
+               switch (ch = string[ci]) {
                case '*':
-                       if (ci + 1 < size && string[ci + 1] == '*') {
-                               if (ci + 2 < size && string[ci + 2] == '*') {
+               case '_':
+                       if (ci + 1 < size && string[ci + 1] == ch) {
+                               printf("ci: %d && ci + 1 == %c\n", ci, ch);
+                               if (ci + 2 < size && string[ci + 2] == ch) {
+                                       printf("ci: %d && ci + 2 == %c\n", ci, ch);
                                        is_bold = !is_bold;
                                        is_italic = !is_italic;
                                        result[ci] = CH_IGNORE;
@@ -2737,25 +2742,10 @@ void
                                ci += 2;
                                break;
                        }
-                       is_italic = !is_italic;
-                       result[ci] = CH_IGNORE;
-                       ci++;
-                       break;
-               case '_':
-                       if (ci + 1 < size && string[ci + 1] == '_') {
-                               if (ci + 2 < size && string[ci + 2] == '_') {
-                                       is_bold = !is_bold;
-                                       is_italic = !is_italic;
-                                       result[ci] = CH_IGNORE;
-                                       result[ci + 1] = CH_IGNORE;
-                                       result[ci + 2] = CH_IGNORE;
-                                       ci += 3;
-                                       break;
-                               }
-                               is_bold = !is_bold;
-                               result[ci] = CH_IGNORE;
-                               result[ci + 1] = CH_IGNORE;
-                               ci += 2;
+                       // Try to avoid flagging a single underscore in the middle of a word.
+                       if (ch == '_' && ci > 0 && string[ci - 1] != ' ' && string[ci + 1] != ' ' && string[ci - 1] != ch && string[ci + 1] != ch) {
+                               result[ci] = CH_REGULAR;
+                               ci++;
                                break;
                        }
                        is_italic = !is_italic;
@@ -2902,7 +2892,7 @@ int
        }
        // Calculate the maximum number of lines we may have to deal with
        unsigned int num_lines = (unsigned int)(viewHeight / (uint32_t)max_height);
-
+
        // And allocate the memory for it...
        lines = calloc(num_lines, sizeof(FBInkOTLine));
 
@@ -2921,7 +2911,7 @@ int
        LOG("Found linebreaks!");
 
        // Parse our string for formatting, if requested
-       fmt_buff = calloc(str_len_bytes, sizeof(char));
+       fmt_buff = calloc(str_len_bytes + 1, sizeof(char));
        if (!fmt_buff) {
                rv = ERRCODE(EXIT_FAILURE);
                goto cleanup;
@@ -3064,7 +3054,7 @@ int
        unsigned char *lnPtr, *glPtr = NULL;
        unsigned short start_x, start_y;
        // stb_truetype renders glyphs with color inverted to what our blitting functions expect
-       unsigned char invert = 0xff;
+       unsigned char invert = cfg->is_inverted ? 0x00 : 0xFF;
        // Render!
        for (line = 0; lines[line].line_used; line++) {
                printf("Line # %d\n", line);
diff --git a/fbink.h b/fbink.h
index 3777e19..8dbe3d0 100644
--- a/fbink.h
+++ b/fbink.h
@@ -198,6 +198,7 @@ typedef struct {
        } margins;
        bool is_centered;         // Horizontal text centering
        bool is_formatted;                // Is string "formatted"? Bold/Italic support only, markdown like syntax
+       bool is_inverted;
 } FBInkOTConfig;
 
 // NOTE: Unless otherwise specified,
@@ -271,7 +272,7 @@ FBINK_API int fbink_printf(int fbfd, const FBInkConfig* fbink_config, const char
     __attribute__((format(printf, 3, 4)));
 
 // Print a string using an OpenType font. Note the caller MUST init with fbink_init_ot() FIRST.
-// This function uses margins (as whole number percentages) instead of rows/columns for 
+// This function uses margins (as whole number percentages) instead of rows/columns for
 // positioning and setting the printable area.
 // Returns -(ERANGE) if the provided margins are out of range, or sum to < 100%
 // Returns -(ENOSYS) if compiled with MINIMAL
diff --git a/fbink_cmd.c b/fbink_cmd.c
index fabf953..6eb5df1 100644
--- a/fbink_cmd.c
+++ b/fbink_cmd.c
@@ -690,11 +690,23 @@ int
                                    fbink_config.fontname,
                                    fbink_config.fontmult);
                        }
+                       // TTF!
+                       FBInkOTConfig cfg = { 0 };
+                       cfg.is_formatted = true;
+                       cfg.margins.top = 5;
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Bold.ttf", FNT_BOLD);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-BoldItalic.ttf", FNT_BOLD_ITALIC);
+                       fbink_print_ot(fbfd, string, &cfg);
+                       fbink_free_ot_fonts();
+                       /*
                        if ((linecount = fbink_print(fbfd, string, &fbink_config)) < 0) {
                                fprintf(stderr, "Failed to print that string!\n");
                                rv = ERRCODE(EXIT_FAILURE);
                                goto cleanup;
                        }
+                       */
                        // NOTE: Don't clobber previous entries if multiple strings were passed...
                        //       We make sure to trust print's return value,
                        //       because it knows how much space it already took up ;).
```

NiLuJe:
And that's it for now ;).

I you have a few test strings that crashes, i'm happy to try that on my end. (I'll probably need to know the font/sizes/margins you were using, though).

NiLuJe:
And with a private stbtt implementation, which may allow GCC some more optimization opportunities...

```diff
diff --git a/fbink.c b/fbink.c
index 36f32eb..7439800 100644
--- a/fbink.c
+++ b/fbink.c
@@ -66,6 +66,8 @@
 // stb_truetype needs maths, and so do we to round to the nearest pixel
 #      include <math.h>
 #      define STB_TRUETYPE_IMPLEMENTATION
+// Make it private, we don't need it anywhere else
+#      define STBTT_STATIC
 // stb_truetype is.... noisy
 #      pragma GCC diagnostic push
 #      pragma GCC diagnostic ignored "-Wunknown-pragmas"
@@ -2114,8 +2116,8 @@ int
 }
 
 // Free an individual OpenType font structure
-void* 
-       free_ot_font(stbtt_fontinfo* font_info) 
+void*
+       free_ot_font(stbtt_fontinfo* font_info)
 {
        if (font_info) {
                free(font_info->data); // This is the font data we loaded
@@ -2712,17 +2714,22 @@ int
 // This is **bold** text.
 // This is ***bold italic*** text.
 // As well as their underscore equivalents
-void 
+void
        parse_simple_md(char* string, int size, unsigned char* result)
 {
        int ci = 0;
+       char ch;
        bool is_italic = false;
        bool is_bold = false;
        while (ci < size) {
-               switch (string[ci]) {
+               printf("ci: %d (< %d) is %c\n", ci, size, string[ci]);
+               switch (ch = string[ci]) {
                case '*':
-                       if (ci + 1 < size && string[ci + 1] == '*') {
-                               if (ci + 2 < size && string[ci + 2] == '*') {
+               case '_':
+                       if (ci + 1 < size && string[ci + 1] == ch) {
+                               printf("ci: %d && ci + 1 == %c\n", ci, ch);
+                               if (ci + 2 < size && string[ci + 2] == ch) {
+                                       printf("ci: %d && ci + 2 == %c\n", ci, ch);
                                        is_bold = !is_bold;
                                        is_italic = !is_italic;
                                        result[ci] = CH_IGNORE;
@@ -2737,25 +2744,10 @@ void
                                ci += 2;
                                break;
                        }
-                       is_italic = !is_italic;
-                       result[ci] = CH_IGNORE;
-                       ci++;
-                       break;
-               case '_':
-                       if (ci + 1 < size && string[ci + 1] == '_') {
-                               if (ci + 2 < size && string[ci + 2] == '_') {
-                                       is_bold = !is_bold;
-                                       is_italic = !is_italic;
-                                       result[ci] = CH_IGNORE;
-                                       result[ci + 1] = CH_IGNORE;
-                                       result[ci + 2] = CH_IGNORE;
-                                       ci += 3;
-                                       break;
-                               }
-                               is_bold = !is_bold;
-                               result[ci] = CH_IGNORE;
-                               result[ci + 1] = CH_IGNORE;
-                               ci += 2;
+                       // Try to avoid flagging a single underscore in the middle of a word.
+                       if (ch == '_' && ci > 0 && string[ci - 1] != ' ' && string[ci + 1] != ' ') {
+                               result[ci] = CH_REGULAR;
+                               ci++;
                                break;
                        }
                        is_italic = !is_italic;
@@ -2788,6 +2780,11 @@ int
 #      pragma GCC diagnostic ignored "-Wconversion"
 #      pragma GCC diagnostic ignored "-Wbad-function-cast"
 
+       // Abort if we were passed an empty string
+       if (! *string) {
+               return ERRCODE(EXIT_FAILURE);
+       }
+
        // Has fbink_init_ot() been called yet?
        if (!otInit) {
                return ERRCODE(ENODATA);
@@ -2902,7 +2899,7 @@ int
        }
        // Calculate the maximum number of lines we may have to deal with
        unsigned int num_lines = (unsigned int)(viewHeight / (uint32_t)max_height);
-
+
        // And allocate the memory for it...
        lines = calloc(num_lines, sizeof(FBInkOTLine));
 
@@ -2921,7 +2918,7 @@ int
        LOG("Found linebreaks!");
 
        // Parse our string for formatting, if requested
-       fmt_buff = calloc(str_len_bytes, sizeof(char));
+       fmt_buff = calloc(str_len_bytes + 1, sizeof(char));
        if (!fmt_buff) {
                rv = ERRCODE(EXIT_FAILURE);
                goto cleanup;
@@ -3064,7 +3061,7 @@ int
        unsigned char *lnPtr, *glPtr = NULL;
        unsigned short start_x, start_y;
        // stb_truetype renders glyphs with color inverted to what our blitting functions expect
-       unsigned char invert = 0xff;
+       unsigned char invert = cfg->is_inverted ? 0x00 : 0xFF;
        // Render!
        for (line = 0; lines[line].line_used; line++) {
                printf("Line # %d\n", line);
diff --git a/fbink.h b/fbink.h
index 3777e19..8dbe3d0 100644
--- a/fbink.h
+++ b/fbink.h
@@ -198,6 +198,7 @@ typedef struct {
        } margins;
        bool is_centered;         // Horizontal text centering
        bool is_formatted;                // Is string "formatted"? Bold/Italic support only, markdown like syntax
+       bool is_inverted;
 } FBInkOTConfig;
 
 // NOTE: Unless otherwise specified,
@@ -271,7 +272,7 @@ FBINK_API int fbink_printf(int fbfd, const FBInkConfig* fbink_config, const char
     __attribute__((format(printf, 3, 4)));
 
 // Print a string using an OpenType font. Note the caller MUST init with fbink_init_ot() FIRST.
-// This function uses margins (as whole number percentages) instead of rows/columns for 
+// This function uses margins (as whole number percentages) instead of rows/columns for
 // positioning and setting the printable area.
 // Returns -(ERANGE) if the provided margins are out of range, or sum to < 100%
 // Returns -(ENOSYS) if compiled with MINIMAL
diff --git a/fbink_cmd.c b/fbink_cmd.c
index fabf953..6eb5df1 100644
--- a/fbink_cmd.c
+++ b/fbink_cmd.c
@@ -690,11 +690,23 @@ int
                                    fbink_config.fontname,
                                    fbink_config.fontmult);
                        }
+                       // TTF!
+                       FBInkOTConfig cfg = { 0 };
+                       cfg.is_formatted = true;
+                       cfg.margins.top = 5;
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Bold.ttf", FNT_BOLD);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-BoldItalic.ttf", FNT_BOLD_ITALIC);
+                       fbink_print_ot(fbfd, string, &cfg);
+                       fbink_free_ot_fonts();
+                       /*
                        if ((linecount = fbink_print(fbfd, string, &fbink_config)) < 0) {
                                fprintf(stderr, "Failed to print that string!\n");
                                rv = ERRCODE(EXIT_FAILURE);
                                goto cleanup;
                        }
+                       */
                        // NOTE: Don't clobber previous entries if multiple strings were passed...
                        //       We make sure to trust print's return value,
                        //       because it knows how much space it already took up ;).
diff --git a/fbink_types.h b/fbink_types.h
index 89d1c5f..42187d2 100644
--- a/fbink_types.h
+++ b/fbink_types.h
@@ -23,6 +23,9 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+// NOTE: We need to import stbtt early because we depend on stbtt_fontinfo here
+//       We'll want it as static/private, so do that here, because we're importing it earlier than fbink.c
+#define STBTT_STATIC
 #include "stb/stb_truetype.h"
 
 // List of flags for device or screen-specific quirks...
```

NiLuJe:
Okay, ungluing myself from the screen, for real this time ;).

NiLuJe:
I'm not sure if that might be an issue here, but I've had weird things happening with libu8 skipping over a single terminating NULL because something looked juicy enough after the end of the buffer ;).
I 'fixed' that by making my buffers terminated by a "wide" NULL (i.e., 4 bytes).

NiLuJe:
And on the topic of possibly unrelated stuff, here's what's Clang's static analyzer had to say:

```
fbink.c:2901:53: warning: Division by zero
        unsigned int num_lines = (unsigned int)(viewHeight / (uint32_t)max_height);
                                                ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
fbink.c:2921:13: warning: Result of 'calloc' is converted to a pointer of type 'unsigned char', which is incompatible with sizeof operand type 'char'
        fmt_buff = calloc(str_len_bytes + 1, sizeof(char));
                   ^~~~~~                    ~~~~~~~~~~~~
fbink.c:2966:9: warning: Assigned value is garbage or undefined
                                        sf = itMetrics.sf;
                                           ^ ~~~~~~~~~~~~
fbink.c:3048:14: warning: Result of 'calloc' is converted to a pointer of type 'unsigned char', which is incompatible with sizeof operand type 'char'
        line_buff = calloc(max_lw * font_size_px, sizeof(char));
                    ^~~~~~                        ~~~~~~~~~~~~
fbink.c:3050:15: warning: Result of 'calloc' is converted to a pointer of type 'unsigned char', which is incompatible with sizeof operand type 'char'
        glyph_buff = calloc(font_size_px * font_size_px * 2, sizeof(char));
                     ^~~~~~                                  ~~~~~~~~~~~~
fbink.c:3084:9: warning: Assigned value is garbage or undefined
                                        sf = itMetrics.sf;
```
I tend to take all of that with a grain of salt, but, who knows, sometimes it does catches things ;).

shermp:
Thanks for the patch. That parsing function truly was quick and dirty, thanks for improving it!

I seem to have lost my test string that was causing a segfault, but after sleeping on it, I may have an idea. it could be a matter of a very slight mismatch between the measuring and rendering stage. To counter this, I currently truncate the glyph width if it exceeds the maximum width. The segfault happened within stb_truetype, so I suspect it may not be happy with truncating small glyphs like '.'.

What I may do instead, is make the measuring maximum line width slightly smaller by a few pixels to ensure that we hopefully never run into the situation of the glyph width exceeding the allowed width. I shall then make it error out if this happens.

shermp:
That's it from me today. Should hopefully no longer segfault, and I've tidied up, added error message where appropriate etc.

Still having a measurement mismatch, which becomes fairly obvious at large font sizes. I'm testing 36 pt at the moment. I'll be a lot happier once (if) I figure out the cause. Not adding any new features until that issue is solved, or a decision is made to say it's "good enough". Knowing my luck, I'll have made an extremely silly simple boo-boo that takes me days of debuggers and print statements to spot what should have been blindingly obvious (that has happened to me before)

NiLuJe:
Heh ;).

I'll try taking a look at it later tonight (somewhat later, because family dinner today ;)).

shermp:
So, just had another look at this issue with a fresh head.

Looks like the measurement issues aren't as bad as I feared. I ran two prints, then lined up the outputs side-by-side, and the measurements are identical... except for the fact there are a couple of extra ones in the measurement phase compared to rendering.

So, if I can figure out the whys of that, then I think I will be fairly confident of all the rendering stuff.

shermp:
Doh... Those extra measurements aren't a problem at all. 🤦

Of course the measurement phase may try more characters that ultimately won't fit on the current line.

Told you it would be something completely obvious :p

NiLuJe:
Okay, back home, have a bit of time :)

I'm seeing segfaults at higher font sizes (I'm testing w/ 36pt right now) and/or massive jumps in LW, leading to unprinted stuff, going to dig into it ;).

It segfaults during `lnPtr[k] = glPtr[k];`, which makes me think of memory management related issues (buffer overflow? corruption?)

shermp:
Note that the following code is *supposed* to prevent that underflow
```
if (curr_point.x == 0 && lsb < 0) {
				curr_point.x += (int)roundf(sf * abs(lsb));
			}
```
Does it not?

Because, if x0 is negative, then so should lsb

NiLuJe:
Also switched from roundf to lroundf, to avoid a float truncation, not sure if that changes anything meaningful, though ;).

```diff
diff --git a/fbink.c b/fbink.c
index e407bbe..93af9d5 100644
--- a/fbink.c
+++ b/fbink.c
@@ -2879,8 +2879,8 @@ int
        if (otFonts.otRegular) {
                rgMetrics.sf = stbtt_ScaleForPixelHeight(otFonts.otRegular, (float)font_size_px);
                stbtt_GetFontVMetrics(otFonts.otRegular, &rgMetrics.asc, &rgMetrics.desc, &rgMetrics.lg);
-               rgMetrics.baseline = (int)roundf(rgMetrics.sf * rgMetrics.asc);
-               int height = (int)roundf(rgMetrics.sf * (rgMetrics.asc - rgMetrics.desc + rgMetrics.lg));
+               rgMetrics.baseline = (int) lroundf(rgMetrics.sf * rgMetrics.asc);
+               int height = (int) lroundf(rgMetrics.sf * (rgMetrics.asc - rgMetrics.desc + rgMetrics.lg));
                if (height > max_font_height) {
                        max_font_height = height;
                }
@@ -2896,8 +2896,8 @@ int
        if (otFonts.otItalic) {
                itMetrics.sf = stbtt_ScaleForPixelHeight(otFonts.otItalic, (float)font_size_px);
                stbtt_GetFontVMetrics(otFonts.otRegular, &itMetrics.asc, &itMetrics.desc, &itMetrics.lg);
-               itMetrics.baseline = (int)roundf(itMetrics.sf * itMetrics.asc);
-               int height = (int)roundf(itMetrics.sf * (itMetrics.asc - itMetrics.desc + itMetrics.lg));
+               itMetrics.baseline = (int) lroundf(itMetrics.sf * itMetrics.asc);
+               int height = (int) lroundf(itMetrics.sf * (itMetrics.asc - itMetrics.desc + itMetrics.lg));
                if (height > max_font_height) {
                        max_font_height = height;
                }
@@ -2913,8 +2913,8 @@ int
        if (otFonts.otBold) {
                bdMetrics.sf = stbtt_ScaleForPixelHeight(otFonts.otBold, (float)font_size_px);
                stbtt_GetFontVMetrics(otFonts.otRegular, &bdMetrics.asc, &bdMetrics.desc, &bdMetrics.lg);
-               bdMetrics.baseline = (int)roundf(bdMetrics.sf * bdMetrics.asc);
-               int height = (int)roundf(bdMetrics.sf * (bdMetrics.asc - bdMetrics.desc + bdMetrics.lg));
+               bdMetrics.baseline = (int)lroundf(bdMetrics.sf * bdMetrics.asc);
+               int height = (int)lroundf(bdMetrics.sf * (bdMetrics.asc - bdMetrics.desc + bdMetrics.lg));
                if (height > max_font_height) {
                        max_font_height = height;
                }
@@ -2930,8 +2930,8 @@ int
        if (otFonts.otBoldItalic) {
                bditMetrics.sf = stbtt_ScaleForPixelHeight(otFonts.otBoldItalic, (float)font_size_px);
                stbtt_GetFontVMetrics(otFonts.otRegular, &bditMetrics.asc, &bditMetrics.desc, &bditMetrics.lg);
-               bditMetrics.baseline = (int)roundf(bditMetrics.sf * bditMetrics.asc);
-               int height = (int)roundf(bditMetrics.sf * (bditMetrics.asc - bditMetrics.desc + bditMetrics.lg));
+               bditMetrics.baseline = (int)lroundf(bditMetrics.sf * bditMetrics.asc);
+               int height = (int)lroundf(bditMetrics.sf * (bditMetrics.asc - bditMetrics.desc + bditMetrics.lg));
                if (height > max_font_height) {
                        max_font_height = height;
                }
@@ -2961,7 +2961,7 @@ int
        // Calculate the maximum number of lines we may have to deal with
        unsigned int print_height = area.br.y - area.tl.y;
        unsigned int num_lines = print_height / max_font_height;
-
+
        // And allocate the memory for it...
        lines = calloc(num_lines, sizeof(FBInkOTLine));
 
@@ -3060,11 +3060,11 @@ int
                        // Note, these metrics are unscaled, we need to use our previously
                        // obtained scale factor (sf) to get the metrics as pixels
                        stbtt_GetCodepointHMetrics(curr_font, (int)c, &adv, &lsb);
-
+
                        // If starting a line with a character that has a negative lsb
                        // eg 'j', move our start point a little.
                        if (curr_x == 0 && lsb < 0) {
-                               curr_x += (int)roundf(sf * abs(lsb));
+                               curr_x += (int)lroundf(sf * abs(lsb));
                        }
                        stbtt_GetCodepointBitmapBox(curr_font, c, sf, sf, &x0, &y0, &x1, &y1);
                        gw = x1 - x0;
@@ -3110,12 +3110,12 @@ int
                                        break;
                                }
                        }
-                       curr_x += (int)roundf(sf * adv);
+                       curr_x += (int)lroundf(sf * adv);
                        // Adjust our x position for kerning, because we can :)
                        if (string[c_index + 1]) {
                                tmp_c_index = c_index;
                                uint32_t c2 = u8_nextchar(string, &tmp_c_index);
-                               curr_x += (int)roundf(sf * stbtt_GetCodepointKernAdvance(curr_font, c, c2));
+                               curr_x += (int)lroundf(sf * stbtt_GetCodepointKernAdvance(curr_font, c, c2));
                        }
                }
                // We've run out of string! This is our last line.
@@ -3137,7 +3137,7 @@ int
        // centering if required.
        line_buff = calloc(max_lw * font_size_px, sizeof(unsigned char));
        // We also don't want to be creating a new buffer for every glyph
-       glyph_buff = calloc(font_size_px * font_size_px * 2, sizeof(unsigned char));
+       glyph_buff = calloc(font_size_px * font_size_px * 8, sizeof(unsigned char));
        if (!line_buff || !glyph_buff) {
                ELOG("[FBInk] Line or glyph buffers could not be allocated");
                rv = ERRCODE(EXIT_FAILURE);
@@ -3151,7 +3151,7 @@ int
        unsigned char *lnPtr, *glPtr = NULL;
        unsigned short start_x, start_y;
        // stb_truetype renders glyphs with color inverted to what our blitting functions expect
-    unsigned char invert = cfg->is_inverted ? 0x00 : 0xFF;
+       unsigned char invert = cfg->is_inverted ? 0x00 : 0xFF;
        // Render!
        for (line = 0; line < num_lines; line++) {
                if (!lines[line].line_used) {
@@ -3198,21 +3198,22 @@ int
                        c = u8_nextchar(string, &ci);
                        stbtt_GetCodepointHMetrics(curr_font, c, &adv, &lsb);
                        if (curr_point.x == 0 && lsb < 0) {
-                               curr_point.x += (int)roundf(sf * abs(lsb));
+                               curr_point.x += (unsigned short int) lroundf(sf * abs(lsb));
                        }
                        stbtt_GetCodepointBitmapBox(curr_font, c, sf, sf, &x0, &y0, &x1, &y1);
                        gw = x1 - x0;
                        gh = y1 - y0;
-                       ins_point.x = curr_point.x + x0;
+                       ins_point.x = MAX(0, curr_point.x + x0);
                        ins_point.y += y0;
+                       printf("gw: %d & gh: %d for c: U+%04X @ ins_point (%hu, %hu) & curr_point (%hu, %hu) / x0: %d y0: %d x1: %d y1: %d\n", gw, gh, c, ins_point.x, ins_point.y, curr_point.x, curr_point.y, x0, y0, x1, y1);
                        // We only increase the lw if glyph not a space This hopefully prevent trailing
                        // spaces from being printed on a line.
-                       if (gw) {
+                       if (gw > 0) {
                                lw = ins_point.x + gw;
                        } else {
                                lw = ins_point.x;
                        }
-                       printf("Current Rendered LW: %d  Line# %d\n", lw, line);
+                       printf("Current Rendered LW: %u  Line# %u\n", lw, line);
                        // Just in case our arithmetic was off by a pixel or two...
                        // Note that we are deliberately using a slightly shorter line
                        // width during the measurement phase, so this should not happen.
@@ -3220,11 +3221,11 @@ int
                        // bounding box, to avoid the possiblity of stb_truetype segfaulting.
                        if (lw > max_lw) {
                                ELOG("[FBInk] Max allowed line width exceeded");
-                               ELOG("[FBInk] Curr LW: %d   Max Allowed: %d", lw, max_lw);
+                               ELOG("[FBInk] Curr LW: %u   Max Allowed: %hu", lw, max_lw);
                                rv = ERRCODE(EXIT_FAILURE);
                                goto cleanup;
                        }
-                       if (gw) {
+                       if (gw > 0) {
                                // Because the stbtt_MakeCodepointBitmap documentation is a bit vague on this
                                // point, the parameter 'out_stride' should be the width of the surface in our
                                // buffer. It's designed so that the glyph can be rendered directly to a screen buffer.
@@ -3247,11 +3248,11 @@ int
                                        lnPtr += max_lw;
                                }
                        }
-                       curr_point.x += (int)roundf(sf * adv);
+                       curr_point.x += (unsigned short int) lroundf(sf * adv);
                        if (ci < lines[line].endCharIndex) {
                                unsigned int tmp_i = ci;
                                tmp_c = u8_nextchar(string, &tmp_i);
-                               curr_point.x += (int)roundf(sf * stbtt_GetCodepointKernAdvance(curr_font, c, tmp_c));
+                               curr_point.x += (unsigned short int) lroundf(sf * stbtt_GetCodepointKernAdvance(curr_font, c, tmp_c));
                        }
                        ins_point.y = baseline;
                }
@@ -3274,7 +3275,7 @@ int
                        paint_point.x = start_x;
                        paint_point.y++;
                }
-               paint_point.y += (int)(sf * lg);
+               paint_point.y += (unsigned short int) lroundf(sf * lg);
                paint_point.x = area.tl.x;
                // Woohoo, it's in our framebuffer! Let's refresh the screen.
                struct mxcfb_rect region = { 0 };
diff --git a/fbink_cmd.c b/fbink_cmd.c
index bbf8c0b..cc7e8a1 100644
--- a/fbink_cmd.c
+++ b/fbink_cmd.c
@@ -694,6 +694,7 @@ int
                        FBInkOTConfig cfg = { 0 };
                        cfg.is_formatted = true;
                        cfg.margins.top = 5;
+                       cfg.size_pt = 36;
                        fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
                        fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
                        fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Bold.ttf", FNT_BOLD);
diff --git a/fbink_internal.h b/fbink_internal.h
index 0a17bd9..5cc8750 100644
--- a/fbink_internal.h
+++ b/fbink_internal.h
@@ -36,6 +36,9 @@
 #      ifndef FBINK_WITH_IMAGE
 #              define FBINK_WITH_IMAGE
 #      endif
+#      ifndef FBINK_WITH_OPENTYPE
+#              define FBINK_WITH_OPENTYPE
+#      endif
 // Connect button scanning is Kobo specific
 #      ifndef FBINK_FOR_KINDLE
 #              ifndef FBINK_FOR_CERVANTES
```

NiLuJe:
I'm getting 'lsb = 7' for my S @ 36pt

shermp:
Bugger. That explains it. Will need to look into that! Maybe reorder some things, and use x0 instead of lsb

NiLuJe:
Deleted my first comment, I was looking at the results from the wrong test string ;).

`gw: 72 & gh: 96 for c: U+0053 @ ins_point (0, 12) & curr_point (0, 106) / x0: -1 y0: -94 x1: 71 y1: 2 / lsb: -2`
`gw: 82 & gh: 101 for c: U+0070 @ ins_point (63, 35) & curr_point (73, 106) / x0: -10 y0: -71 x1: 72 y1: 30 / lsb: -74`

So, yes, lsb is negative ;).

shermp:
I see the problem. lsb is so small, that when scaled, it probably rounds down to zero. Need to change that to round up.

Or just use x0 to be extra safe...

NiLuJe:
Small correction: that was a Bold Italic 'S'.

Note that regular 'Sp' segfaults anyway, though ;).

NiLuJe:
A quick printf hack indeed seems to yield `scaled lsb: 0` ;).

shermp:
Umm, yeah, about that 'Sp' segfault. I set the glyph buffer size to twice the glyph height. Forgetting about things like ligatures, and other compound glyphs. Oops.

shermp:
I shall apply your patch and sort out that lsb matter. Thank you very much for the testing.

NiLuJe:
Happy to, thanks for working on this ;).

shermp:
Hopefully that improves matters.

Note, I applied your fixes manually because git threw a hissy fit, and I didn't yet feel like rebasing on master again.

I've used a slightly different solution for the underflow issue. The point is not to cut the glyph off at all, and the solution should be equivalent to what I was doing with lsb (had that worked correctly). In short, curr_point.x should no longer go below zero.

I tested with an italic 'J', but I don't have bookerly with that 'S'...

NiLuJe:
I'll check how that behaves w/ Bookerly, thanks ;).

I *think* I can push to this branch, since it's an open PR, if it can save you some git weirdness...

NiLuJe:
:+1:

```
[FBInk] 0 lines to be printed
Line # 0
gw: 72 & gh: 96 for c: U+0053 @ ins_point (0, 12) & curr_point (1, 106) / x0: -1 y0: -94 x1: 71 y1: 2 / lsb: -2
Current Rendered LW: 72  Line# 0
gw: 82 & gh: 101 for c: U+0070 @ ins_point (64, 35) & curr_point (74, 106) / x0: -10 y0: -71 x1: 72 y1: 30 / lsb: -74
Current Rendered LW: 146  Line# 0
```

NiLuJe:
Note that if I jump to 64pt, the segfaults becomes more vicious:

```
gw: 127 & gh: 171 for c: U+0053 @ ins_point (0, 20) & curr_point (1, 188) / x0: -1 y0: -168 x1: 126 y1: 3 / lsb: -2
Current Rendered LW: 127  Line# 0
gw: 146 & gh: 180 for c: U+0070 @ ins_point (113, 62) & curr_point (131, 188) / x0: -18 y0: -126 x1: 128 y1: 54 / lsb: -74
Current Rendered LW: 259  Line# 0
*** Error in `./fbink': munmap_chunk(): invalid pointer: 0x2b81c008 ***
[1]    4637 abort      ./fbink '***Sp***'
```

shermp:
Hmm, I'm not actually using fbink_cmd to test. Can you let me know the test string and fbink arguments you are using to see if I can replicate this?

I shall put bookerly on my device (I think I have a version of it somewhere) and test.

NiLuJe:
It's literally `./fbink 'Sp'` with the current hacked fbink_cmd ;).

And, yeah, needed an access token because 2FA, but push is working ;).

NiLuJe:
I'm vaguely hoping that won't be specific to *my* Bookerly version, because I may have tweaked it at some point...

Give me a shout if you can't reproduce, and I'll send it to you.

NiLuJe:
Doubling the size of `line_buff` appears to 'fix' the 64pt crash, at a glance.

NiLuJe:
Yeah, I'll bundle my Bookerly version, that'll probably be easier to replicate.

shermp:
I haven't been able to replicate with my standard version of Bookerly. Just tried ./fbink 'Sp' at 128 points with no issue

NiLuJe:
[Here goes](http://files.ak-team.com/niluje/tmp/fbink_kablooey.tar.gz) ;).

NiLuJe:
Yeah, no issues either with a vanilla Bookerly on my end, even with the default buffer sizes, and large font sizes.

FWIW, the tweaks involved are the usual Panose crap to make the kepub renderer happy.

shermp:
Ok, managed to replicate at 64pt, with your version of Bookerly.

128pt works fine BTW

shermp:
I would speculate there is a mismatch between the height I calculate from the metrics (which is used to calculate the buffer size) and the heights getting from the bounding boxes.

Let's see what I can do about it.

NiLuJe:
FWIW, at 128pt:

```
Tweaked:

font_size_px: 471
[FBInk] Unformatted text defaulting to regular font
Current Measured LW: 236  Line# 0
Current Measured LW: 529  Line# 0
Current Measured LW: 830  Line# 0
[FBInk] 0 lines to be printed
Line # 0
gw: 214 & gh: 341 for c: U+0053 @ ins_point (22, 46) & curr_point (0, 378) / x0: 22 y0: -332 x1: 236 y1: 9 / lsb: 47
Current Rendered LW: 236  Line# 0
gw: 267 & gh: 357 for c: U+0070 @ ins_point (262, 128) & curr_point (259, 378) / x0: 3 y0: -250 x1: 270 y1: 107 / lsb: 7
Current Rendered LW: 529  Line# 0
gw: 279 & gh: 352 for c: U+0079 @ ins_point (551, 137) & curr_point (548, 378) / x0: 3 y0: -241 x1: 282 y1: 111 / lsb: 8
Current Rendered LW: 830  Line# 0

Vanilla:

font_size_px: 471
[FBInk] Unformatted text defaulting to regular font
Current Measured LW: 179  Line# 0
Current Measured LW: 402  Line# 0
Current Measured LW: 630  Line# 0
[FBInk] 0 lines to be printed
Line # 0
gw: 163 & gh: 259 for c: U+0053 @ ins_point (16, 126) & curr_point (0, 378) / x0: 16 y0: -252 x1: 179 y1: 7 / lsb: 47
Current Rendered LW: 179  Line# 0
gw: 203 & gh: 271 for c: U+0070 @ ins_point (199, 188) & curr_point (197, 378) / x0: 2 y0: -190 x1: 205 y1: 81 / lsb: 7
Current Rendered LW: 402  Line# 0
gw: 212 & gh: 267 for c: U+0079 @ ins_point (418, 195) & curr_point (416, 378) / x0: 2 y0: -183 x1: 214 y1: 84 / lsb: 8
Current Rendered LW: 630  Line# 0
```

NiLuJe:
Here be dragons, because I'm starting to fall asleep:

Doesn't the fact that stbtt_MakeGlyphBitmapSubpixel (i.e., stbtt_MakeCodepointBitmap) calls stbtt_GetGlyphBitmapBoxSubpixel (i.e., stbtt_GetCodepointBitmapBox) again could possibly be screwing things up?

shermp:
So I think I figured out what the problem was.

Turns out the metrics on your Bookerly font are fibbing. More specifically, the glyphs rendered can descend lower than what the metrics state. And all my measurements were done assuming the metrics are canon.

What I believe was happening is that because the line buffer was too small (not tall enough), it was overflowing into the glyph buffer, and then segfaulting when attempting to free the glyph buffer.

I've completely redone the vertical measuring. The font metrics are now treated as suggestions, and during the glyph measurement phase, I update the max line height if necessary, before it is allocated. This seems to be working alright, although I've changed a fair amount, so new bugs could have been introduced.

Note, I have seen instances of minor clipping with Bookerly at 64 pt with this new code path. While it works, I would recommend using a font where the metrics aren't messed up...

shermp:
For those wondering why I'm not over allocating, I'm trying my best to be as miserly as possible with memory usage. Loading up to four fonts into memory is bad enough as it is, especially if each font file is over a megabyte! Thankfully, most fonts are well under 500KB.

(I suppose I could always try mmaping the font files, but I shudder to think of what the performance hit may be...)

NiLuJe:
Yeah, I was going to say I'm fine with clipping on misbehaving fonts, as
long as nothing blows up, so, not a problem ;).

More tests later today, thanks!

NiLuJe:
That appears to do the trick over here, :+1: ;).

I can even shrink the glyph buffer again with no issues:

```diff
diff --git a/fbink.c b/fbink.c
index a810a8f..d706a7f 100644
--- a/fbink.c
+++ b/fbink.c
@@ -2956,7 +2956,7 @@ int
                if (!curr_font) {
                        curr_font = otFonts.otBoldItalic;
                        sf = bditSF;
-                       ELOG("[FBInk] Unformatted text defaulting to bold font");
+                       ELOG("[FBInk] Unformatted text defaulting to bold italic font");
                }
        }
        // If no font was loaded, exit early. We checked earlier, but just in case...
@@ -3167,7 +3167,8 @@ int
        // centering if required.
        line_buff = calloc(max_lw * max_line_height, sizeof(*line_buff));
        // We also don't want to be creating a new buffer for every glyph
-       glyph_buff = calloc(font_size_px * font_size_px * 8, sizeof(*glyph_buff));
+       glyph_buff = calloc(font_size_px * MAX(font_size_px, max_line_height), sizeof(*glyph_buff));
+       printf("max_line_height: %i vs. font_size_px: %u\n", max_line_height, font_size_px);
        if (!line_buff || !glyph_buff) {
                ELOG("[FBInk] Line or glyph buffers could not be allocated");
                rv = ERRCODE(EXIT_FAILURE);
diff --git a/fbink_cmd.c b/fbink_cmd.c
index 9960194..0715bff 100644
--- a/fbink_cmd.c
+++ b/fbink_cmd.c
@@ -694,16 +694,27 @@ int
                        FBInkOTConfig cfg = { 0 };
                        cfg.is_formatted = true;
                        cfg.margins.top = 5;
-                       cfg.size_pt = 18; 
-                       // fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
-                       // fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
-                       // fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Bold.ttf", FNT_BOLD);
-                       // fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-BoldItalic.ttf", FNT_BOLD_ITALIC);
+                       cfg.size_pt = 36;
 
+                       /*
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-Bold.ttf", FNT_BOLD);
+                       fbink_add_ot_font("/mnt/onboard/fonts/Bookerly-BoldItalic.ttf", FNT_BOLD_ITALIC);
+                       */
+
+                       fbink_add_ot_font("/mnt/onboard/.adds/fonts/Bookerly-Regular.ttf", FNT_REGULAR);
+                       fbink_add_ot_font("/mnt/onboard/.adds/fonts/Bookerly-Italic.ttf", FNT_ITALIC);
+                       fbink_add_ot_font("/mnt/onboard/.adds/fonts/Bookerly-Bold.ttf", FNT_BOLD);
+                       fbink_add_ot_font("/mnt/onboard/.adds/fonts/Bookerly-BoldItalic.ttf", FNT_BOLD_ITALIC);
+
+                       /*
                        fbink_add_ot_font("/mnt/onboard/fonts/Alegreya-Regular.ttf", FNT_REGULAR);
                        fbink_add_ot_font("/mnt/onboard/fonts/Alegreya-Italic.ttf", FNT_ITALIC);
                        fbink_add_ot_font("/mnt/onboard/fonts/Alegreya-Bold.ttf", FNT_BOLD);
                        fbink_add_ot_font("/mnt/onboard/fonts/Alegreya-BoldItalic.ttf", FNT_BOLD_ITALIC);
+                       */
+
                        fbink_print_ot(fbfd, string, &cfg);
                        fbink_free_ot_fonts();
                /*
```

(That MAX() might be overkill, if max_line_height happens to always be >= font_size_px by design)).

NiLuJe:
Hmm, might still be a tad optimistic without the * 2, because some glyphs also have a width larger than font_size_px (a regular, capital 'W', in my case) ;).

(... again, with my broke-ass Bookerly :D).

shermp:
Hey, for features like font colours and overlay support etc, would you prefer me to add fields to the OT config struct? Or should I make it that the user can (optionally) add the main FBInk config struct as well?

NiLuJe:
I was wondering that myself when I added the invert stuff... Re-using the main FBInkConfig struct for that sort of stuff probably makes sense, leaving only the OT-specific stuff in the OT one.

shermp:
Yeah, thinking of making it optional (so the user can pass NULL if they don't need it). Then just need to do a NULL check before checking for settings.

Am still thinking of keeping text alignment with the OT config struct, because I don't require a re-init to change alignment.

shermp:
Also, font colour is going to be... interesting.

Basically, the glyphs rendered by stb_truetype are supposed to be interpreted as an 8-bit alpha channel. I'm currently treating it as a simple 8bpp opaque image (except for the lowest (or highest when inverted)  which I treat as fully transparent.

The glyphs are anti-aliased, and so almost all values are used.

So, we have two main options that I can think of.

1. Clip the glyph to the maximum colour. Advantage is simple and quick. Disadvantage is it could introduce glyph artifacts.
2. The proper way of treating the value as an intensity of the desired colour. This is my preferred option if it can be done efficiently. I'm not really up with the maths on this as to how best to do it. My naive approach would be to do something like `(int)(lroundf((intensity / 255.0f) * new_color), but surely there's a better way to do it...

shermp:
The question is, do the performance gains of the 0x00 and 0xff shortcuts make up for the requirement to branch inside a nested loop...? (although, I am already doing that, so hmmm, stuff to think about)

NiLuJe:
I was worried about that (which is partly why I didn't bother with it at first), and, for images, yeah, that was very much worth it ;).

NiLuJe:
I went with 0xFF -> 0 -> blending for images, but if we're being extremely pedantic and worrying about ordering the branches, I'd say 0 -> 0xFF -> blending in this case?

shermp:
Note that I don't think true blending is necessary in this case. If opacity is 0, then I don't overwrite the pixel, otherwise I do. (I'm doing this in my buffer, with a constant background)

shermp:
(I have to do this because characters like `ij` can overlap)

NiLuJe:
Ah, true, I was wondering how that was working when I noticed it did, and then promptly forgot/was too lazy to check the code ;).

shermp:
Right, hopefully nearing feature completion. Have implemented horizontal and vertical alignment, foreground and background colours, fgless, bgless and overlay. 

I haven't bothered with dealing with anti-aliasing for the latter three options however, so expect some minor artifacts. If someone wants to improve that section, please feel free. I'm also cringing a bit at the use of three nearly identical loops in that section too.

From my point of view, the major areas left are dealing with the DPI properly per device, and deciding whether I'm going to convert the margins to pixels from percentages.

This piece of code has turned into a bit of a monster I'm afraid. 702 total LOC and counting so far, just for the print function! Umm, yeah, sorry about that...

shermp:
Here's another interesting dilemma... How to handle fbink_cmd?

The main issue I see, is the fact that a single fbink command would have to open fonts, write a string, close everything. This is fine for a one-off, but it would get very expensive if used in a loop in a shell script for example.

I can think of a couple of different options:

1. Use the above approach anyway and take the performance hit.
2. Don't include OT functionality with fbink_cmd
3. Allow fbink_cmd to run as a daemon, from which the user could send commands via a named pipe.

Thoughts?

NiLuJe:
I was clearly planning to go with 1., via a getsubopt like for image
handling ;).

shermp:
Fine by me. I'll let you handle the fbink_cmd stuff shall I?

NiLuJe:
Sure, no problem!

Just a quick practical decision: do you want me to do in this branch, of after the merge?

Both would *probably* benefit from a rebase before I get to it (especially for the DPI stuff, since that'll be in device_id, which I've been playing with the past few days) ;).

NiLuJe:
Ah, and another random question before I forget: can we shrink the glyph_buff buffer back to font_px * max_h * __2__, or do you want to keep it bigger just in case?

shermp:
Might be better to do so after merge.

As far as glyph buffer is concerned, I may reduce the multiplication factor to four, which should be big enough, and add a check to abort should an extra large glyph ever exceed the value (Probably a good idea regardless).

Just another thought, do we want to provide our own version of the necessary functions from <math.h>, to avoid having to link against libm? stb_truetype allows us to use our own implementations of the functions it requires.

shermp:
Ok, modified the margin handling to now directly use pixel values instead of percentages.

`fbink_print_ot()` now behaves more like `fbink_print()` in that the return value informs the caller of the position to print the next line (in the form of a new top margin value).

NiLuJe:
On a whim, I spent the afternoon reading stuff about the implementations of various stuff from math.h, so, yeaaaah, no, let's not do that :D.

On recent devices, the glibc, despite being terribly old, is still recent enough to have fairly solid implementations of the few things we need, so, I'm okay with that ;).

(Unless you really want to, of course ;p.)

NiLuJe:
I seem to remember realloc being potentially quirky, and the recommended way to deal with it is to assign its return value to a temporary pointer, doing the memory chekc on that one, and only updating the original if it's actually successful ;).

There may even be an example of this somewhere around [here](https://github.com/NiLuJe/FBInk/blob/65afd6f7fa89c9c780b19d73832193d426604723/fbink.c#L3157) ;).

shermp:
Fair enough on the math. 

I've decided I'd better honor is_cleared and is_flashing as well. While I'm doing so, I'm going to change my eink refresh from being per line, to just once for the function call. Should speed up displaying multiple lines hopefully.

NiLuJe:
Yeah, I retro-fitted that in fbink_print in a fairly terrible way, but I think you should have a much easier time of it given the design here ;).

shermp:
Sigh.. that feeling when you made a change, and then when it doesn't work, you add all sorts of print statements to debug... only to discover you forgot to add the config struct to your function call in the test program...

shermp:
There we go. Changed the eink refresh implementation, and I've also implemented support for the is_cleared and is_flashing options.

I think that's about it, apart from the previously discussed DPI matter.

NiLuJe:
:+1:

I'll take a final look at it tomorrow then, and I'll handle the dpi & CLI tweaks once it's merged ;).

Many thanks for working on this!

shermp:
Thanks for the assistance!

I feel a bit bad for opening this PR so early, but your input has been greatly appreciated, and has hopefully resulted in a much more complete and stable solution.

NiLuJe:
As you've probably gathered by now, I'm pretty much in the "throw stuff at the wall, and see what sticks" camp, so, really, not a problem ;).

NiLuJe:
Here we go! :).

Thanks!

shermp:
You're welcome.

Although I just noticed a boo-boo I made that got merged. I forgot to reword the fbink_print_ot() API documentation regarding margins. It is still referring to percentages instead of pixels.

NiLuJe:
Yeah, I noticed, no worries, I'll take care of it ;).

shermp:
Thanks

NiLuJe:
One other thing I just noticed, which I haven't yet looked into: with (at least) valign CENTER,  when re-using the top margin, subsequent print calls won't be edge-to-edge, the second line will start at 3/4 down ;).

shermp:
It's that way by design. Alignment is not to center of the viewport, but rather center to whatever the printable area is, which is defined by the margins. It allows for some very creative positioning.

But now that you've brought it up, I can also see the potential issue with it.

However, as soon as you make multiple print calls for vertical centering, you lose the perfect centering anyway (as previous lines aren't reprinted), so you may as well not use valign, but something like a 50% top margin.

TLDR, vertical centering really only makes sense in the context of a single print call (which honors linefeeds, so you can still do multiple lines in one call, if required)

NiLuJe:
Yep, that makes perfect sense, will just have to jot that down somewhere ;).

shermp:
I feel a bit bad for not doing a better job of cleaning up the type conversions/casting. My apologies for that, and thank you.

Unfortunately, it was the nature of the libraries used that required so much type casting. And in my own code, there was quite a bit of *well what type do I use here...?* going on.

NiLuJe:
Well, to be fair, I do use a fairly psychotic amount of warnings, so, that doesn't help ;).

And yeah, the fact that stb is engineered by a long-time Win32 games dev relying on a really old VS toolchain means no shiny C99 types, and a fine balancing act between memory usage and performance when it comes to the choice of types.

Which leaves us with a few questionable choices to deal with (signed codepoints, really? :D).

TL;DR: What was left was mostly making GCC happy one way or another, don't worry about it ;).

NiLuJe:
@shermp: Quick question, since I'll probably attempt to looking at the AA stuff in esoteric rendering modes over the next few days: What was the reasoning behind inverting the colors if fg < bg? Was that only for the sake of the layer_diff computation, or am I missing something? (I haven't looked into it that much, so I very well may be ;)).

because if it's only about layer_diff, I was thinking that abs(fg -bg) *should* do the trick?

shermp:
Now that I think about it, layer diff could have been made negative and it still would work. The main concern is that when performing the following calculation:
```
lnPtr[k] = (unsigned char) DIV255( (bgcolor + (glPtr[k] * layer_diff)));
```
the fg should be darker than bg if "black on grey/white" or lighter if "white on black/grey"

I was trying to get my head around it, and realized that the two situations were the inverse of each other, hence the inversion. But thinking of it now, a simple negative `layer_diff` where needed would do the same thing. In other words, I over complicated things :p

shermp:
Just some musings on the AA side of things, I've just had an idea.

Currently, each 'line' is prepared with the background colour, and then the glyphs are rendered onto it, baking the final colour information in the process.

What if instead, we treat the line itself as an alpha, and only when painting to the framebuffer do we set the background/foreground colour etc. This would preserve the full AA information in the glyph right up to the final rendering stage.

Thoughts?

NiLuJe:
I *think* that's what I was thinking of, yeah: keeping the alpha coverage map we get from stbtt up until the final rendering stage, where we "just" have to alpha blend it (with the usual 0/0xFF shortcuts to avoid the maths).

NiLuJe:
Okay, started experimenting for real in the ot-blend branch, because trying to figure that out without code & testing was starting to break my brain ^^.

So far, so good... (3c7b0e85caa9f7a0b856bb44e4b6d4ec3dce926b).

Performance is roughly similar, so I'm not even sure I'll bother adding a fast path for B/W||W/B where we could just paint the mask inverted or not...

shermp:
Looks good!

I probably should have taken this approach originally, but I'm still rather a novice at this whole per-pixel image manipulation business...

NiLuJe:
I nearly threw the Kindle 3 out the windows near the end, there... :D.

But apart from the monstrosity that is handling those 4bpp fbs, it went fairly okay, once I realized I'd forgot a multiplication... -_-".

shermp:
Just noticed another potential boo-boo. Where we are painting lines using the loop `for (unsigned int j = 0U; j < font_size_px; j++)`, we probably need to change `font_size_px` to `max_line_height`. Might help avoid any potential clipping with your favorite font...

PR #21, opened 2018-11-02
Allow larger font sizes, and fix UTF-8 iteration for OT support

shermp:
Fixes the issues found [here](https://www.mobileread.com/forums/showpost.php?p=3768773&postcount=45).

I must have been having a brain fade when I implemented the string iteration loop. Hopefully this PR solves that. My test version seems to have worked, as seen [here](https://www.mobileread.com/forums/showpost.php?p=3768849&postcount=56)


NiLuJe:
:+1:

Thanks!

PR #22, opened 2018-11-03
Alpha-blending of AA in OT

NiLuJe:
Using stbtt's alpha coverage mask when we have to...

c.f. the last few comments in #20 

----

I'll let it sit for a while, because fire bad, tree pretty right now. >_<".

But it appeared to behave okay on 32bpp, 8bpp, and 4bpp (*AAAAAAAAAARRRRRRRRRGGGGHHHH*).


NiLuJe:
I refrained from going crazy on the per-bpp branches like in the image codepath, because that was painful enough ;p.

PR #23, opened 2018-11-05
Printf functionality for OT/TT users

shermp:
No great hurry on this, since I missed 1.8.1 (and didn't start working on it untill afterwards anyway)

I thought OT/TT users might like to join in on the printf fun, so I implemented that.

Unfortunately, I couldn't reuse `fbink_printf` as-is, because of the assumptions made regarding buffer lengths. `fbink_printf_ot` mostly copies `fbink_printf`, except for the fact I set the initial buffer size to the input string size, and reallocate if necessary.

I have to do this because there is no easily defined maximum number of characters to print at this stage. We only gain that information during the measurement phase of `fbink_print_ot`.

Tested to be working alright on my Kobo (in English at least), including mixing format flags, and our markdown syntax. `vsnprintf` only cares about the asterisk in the context of a format flag, which then gets removed in the output, so our md parser won't see it. At least, that's how I *think* it works...

### DIFF
--- a/fbink.c
+++ b/fbink.c
@@ -2841,7 +2841,12 @@ int
 	int size = vsnprintf(buffer, buffer_size, fmt, args);
 	// vsnprintf will tell us if it truncated our string to fit the buffer, and by how much. We will need
 	// to realloc if this occurs.
-	if (size >= 0 && (size_t)size >= buffer_size) {
+	// We may as well check for an error from vsnprintf() while we're at it.
+	if (size < 0) {
+		free(buffer);
+		WARN("Could not format string");
+		return ERRCODE(EXIT_FAILURE);
+	} else if ((size_t)size >= buffer_size) {
 		char* tmp_buf = NULL;
 		size_t new_size = (size_t)size + 4U; // Just to be extra safe, use a wide null terminator
 		tmp_buf = realloc(buffer, new_size);
@@ -2858,11 +2863,6 @@ int
 		memset(buffer, '\0', new_size);
 		// Finally, we call vsnprintf() again, this time with the new buffer that should fit our formatted string
 		vsnprintf(buffer, new_size, fmt, args);
-	// We may as well check for an error from vsnprintf() while we're at it.
-	} else if (size < 0) {
-		free(buffer);
-		WARN("Could not format string");
-		return ERRCODE(EXIT_FAILURE);
 	}
 	va_end(args);
 
--- a/fbink.c
+++ b/fbink.c
@@ -2811,6 +2811,70 @@ static void
 }
 #endif    // FBINK_WITH_OPENTYPE
 
+// printf-like wrapper around fbink_print_ot ;).
+int
+	fbink_printf_ot(int fbfd	UNUSED_BY_MINIMAL,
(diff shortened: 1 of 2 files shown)


NiLuJe:
Well, you're not the only one on the "missed opportunity" train, as I've just figured out why Kobo's obfuscated fonts were segfaulting instead of aborting nicely (spoiler: my fault :D). (c.f., d4ae4143ec9fdbf93ffa26e9c7042690149d7e35 & f3591799b14cd9eaac9a154b0ea7e6f7a1ce5663).

----

GCC (Clang supports that, too, IIRC) would *probably* catch a broken fmt string at compile-time (that's what the attribute is for), so, yeah, that looks to be the way to go ;).

NiLuJe:
I'll take a proper look at it after a good night's sleep :).

NiLuJe:
:+1:

Thanks!

shermp:
You're welcome

NiLuJe:
I err, may have reworked the vsnprintf logic, because this made me realize I wasn't a huge fan of the massive alloc in `fbink_printf` either...

Ooops?

shermp:
I noticed your comments re: code duplication.

Is it possible to pass the variadic parameters unchanged to another function? If so, you could create a single vsnprintf function which returns a string buffer, with two wrapper functions that call it. The wrapper functions would then hopefully take care of the compile time check magic you want to keep. Would that work?

Also, great that vsnprintf works with a NULL buffer. I wasn't sure if it did, which is why I performed the initial calloc.

NiLuJe:
I would imagine there's probably a way to do it, because that seems like a common enough scenario, but I'm just not sure ;).

----

Yeah, I missed that vsnprintf trick the first time around, too. After looking around a bit on SO, turns out there was a perfectly fine example in the manpage... Ooops. :D.

NiLuJe:
Hmm, If I understood http://c-faq.com/varargs/handoff.html right, that'd be a no :/.

I can pass the va_list around, but that doesn't help, because that's essentially the bit of code I want to dedup ;p.

NiLuJe:
There's the other approach of only keeping one printf wrapper, and having it accept a function pointer, but, err, that's no fantastically user-friendly ;).

NiLuJe:
Ooh, unless it assumes if it gets an FBInkOTConfig pointer -> OT, otherwise, cell.

Hmm.

NiLuJe:
That'd imply scrapping a bit of the API, but I think I'm the only actual user of fbink_printf in KFMon...

shermp:
Great minds... Just thinking that myself :)

shermp:
Also, if you end up changing that particular API, you won't break go-fbink, as fbink_printf() was never implemented. (CGO doesn't support variable arguments, and Go users can pre-format the string themselves if they need that functionality).

NiLuJe:
I wasn't sure I remembered that right, thanks for the confirmation!

Well then, first, cookies, then back to it :)

NiLuJe:
So far, so good :).

NiLuJe:
And KFMon didn't blow up, victory!

And I only got stuff mildly wrong in the tag description :D.

PR #29, opened 2018-11-23
add a script to build debian packages & use it for Cervantes

pazos:
It relies on `dpkg-deb` on `$PATH`. Intended for cervantes but should work as is for other targets and architectures.

Debian arch names != `uname -m`. More info on https://www.debian.org/ports/

This kind of package won't be accepted by Debian, of course,  but is a nice way to redistribute releases for supported targets :+1:

### DIFF
--- a/tools/do_debian_package.sh
+++ b/tools/do_debian_package.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-# Script to generate debian packages for FBInk. 
+# Script to generate debian packages for FBInk.
 PACKAGE="FBInk"
 AUTHOR="NiLuJe <niluje@gmail.com>"
 DESC_1="FrameBuffer eInker"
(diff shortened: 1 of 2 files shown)


tools/do_debian_package.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# Temp stuff.
+PKG_PATH="${ROOTFS_BASE}/pkg"
+BIN_PATH="${PKG_PATH}/usr/bin"
+DEB_PATH="${PKG_PATH}/DEBIAN"
+CONTROL="${DEB_PATH}/control"
+
+# Get version from git if posible. Fallback to a known version ;p
+# Note: version string must start with a number
+VERSION="$(git describe 2>/dev/null | sed 's/[^0-9]*//' || echo '1.9.0')"
```
NiLuJe: You should be able to rely on the fallback in fbink_internal.h, via something like `grep "define FBINK_VERSION" fbink_internal.h | cut -f2 -d\" | cut -f2 -dv` ;).

pazos:
> I haven't used Debian in over 15 years.

You're a lucky guy. Debian is fine to bootstrap your installation to other computers (thanks to debootstrap/multistrap) but for myself I prefer a distro where I have full control, like Slackware. Sadly  I need to run Ubuntu/Fedora sometimes and have a  couple of VMs ready to cross compile stuff.

NiLuJe:
Yep, Gentoo everywhere \o/.

PR #33, opened 2019-03-28
Dump & Restore feature

### LINKED ISSUE #31: [Idea] Remember (partial) framebuffer state
Hi @NiLuJe , I'm baaack :D

I thought it would be great if one could show an image (or dialog box, or text...), and then when done, restore the original framebuffer state, so that the user isn't presented with the remnants of what we printed when we are done.

I was thinking perhaps it could be implemented as a 2 or three element stack, with each element saving the origin, dimensions, and raw scanlines of the region to be saved.

The user could then call a function like `fbink_save_region()` up to as many times until the stack is full. The could then restore the previous state with a function like `fbink_restore_region()`, again could be called multiple times until the stack is empty.

The main issue I see is memory usage, which is why the stack size would have to be fairly limited.

Thoughts?

NiLuJe:
Basically, what was discussed in #31 (fix #31) ;).

Adds three new API calls:
`fbink_dump`, which dumps the *full* screen
`fbink_region_dump`, which dumps a specific region of the screen, supporting all the same crazy positioning options of the image printing code.
`fbink_restore`, which restores those two kinds of dumps, in the exact same coordinates they were taken from, and errors out if it can't (bitdepth or rotation mismatch between dump & restore time).

This requires a new struct:
`FBInkDump`, which holds a pointer to the FBInk-allocated storage, and the metadata it needs for the safety checks (i.e., coordinates, size, rota, bpp, is it full-screen).
Needs to be zero-initialized, and free'ing is left to the caller. Don't leave the data pointer dangling, otherwise the _dump functions will try to recycle the struct by freeing the data!
This means creating a single struct and keeping it around is perfectly valid, but it will only ever hold a single set of data, that from the latest dump call on it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,6 +363,8 @@ utils: outdir
 	$(STRIP) --strip-unneeded $(OUT_DIR)/rota
 	$(CC) $(CPPFLAGS) $(EXTRA_CPPFLAGS) $(CFLAGS) $(EXTRA_CFLAGS) $(SHARED_CFLAGS) $(LIB_CFLAGS) $(LTO_CFLAGS) $(LDFLAGS) $(EXTRA_LDFLAGS) -o$(OUT_DIR)/fbdepth utils/fbdepth.c
 	$(STRIP) --strip-unneeded $(OUT_DIR)/fbdepth
+	$(CC) $(CPPFLAGS) $(EXTRA_CPPFLAGS) $(CFLAGS) $(EXTRA_CFLAGS) $(SHARED_CFLAGS) $(LIB_CFLAGS) $(LTO_CFLAGS) $(LDFLAGS) $(EXTRA_LDFLAGS) -o$(OUT_DIR)/dump utils/dump.c $(LIBS)
+	$(STRIP) --strip-unneeded $(OUT_DIR)/dump
 
 strip: static
 	$(MAKE) stripbin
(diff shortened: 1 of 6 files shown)


NiLuJe:
Includes a stupid regression fix for Legacy Kindles, and a fast-path when blitting 8bpp images on an 8bpp fb.

Stupid test app for the feature available in utils/dump.c ;).

NiLuJe:
It appears to be handling the usual rotation trickeries, as well as 4bpp fbs.

All in all, that went fairly smoothly, but then, I'm fresh from the same kind of stuff in https://github.com/koreader/koreader-base/pull/878 ;).

NiLuJe:
This pull request **introduces 6 alerts** when merging 3d821f7e2551f62c9d03cafdae7259b01b5cf809 into 18ec76f33982d6b918e0c2fda8828a64d471be10 - [view on LGTM.com](https://lgtm.com/projects/g/NiLuJe/FBInk/rev/pr-37d7c8efab59bb6a59eaa27f9d9a334a011a3738)

**new alerts:**

* 6 for Multiplication result converted to larger type

---

*Comment posted by [LGTM.com](https://lgtm.com)*

NiLuJe:
Up next: I added a `size` field to the struct, which, besides allowing more safety checks, in the unlikely event the framebuffer memory *shrunk* (yeah, that should *never* happen :D), allows one to use it as input for `fbink_print_raw_data`, in the unlikely event you'd want to do that...

I updated the test app with a fairly stupid test for this: taking a region dump @ 32bpp, and restoring it via `fbink_print_raw_data` @ 8bpp, but inverted.
A slightly less tortuous use case would be simply restoring it at different coordinates, for instance.

PR #34, opened 2019-04-18
Image scaling support, thanks to Qt5's QImageScale

NiLuJe:
Ripped out of Qt, trimmed down to the essentials, with added Grayscale (Y8/Y8A) support, and ported back to C.

I blame @Frenzie for the impetus to look into this ;).


Frenzie:
Impressive. :-)

PR #37, opened 2019-09-11
free_ot_font: Double Free

### LINKED ISSUE #36: free_ot_font: Double Free
When calling `fbink_free_ot_fonts()` twice there is a double free.
for example:
```C
fbink_add_ot_font("font.ttf", FNT_REGULAR);
fbink_free_ot_fonts();
...
fbink_add_ot_font("font.ttf", FNT_REGULAR);
fbink_free_ot_fonts();
```
in a long running program.

https://github.com/NiLuJe/FBInk/blob/5642e16cd9859c9e815322f82626232396ecbdc5/fbink.c#L2720
only nulls the local copy of `font_info`
however the global pointer `otFonts.otRegular` is not nulled, which leaves a dangling pointer -> double free.

I am not sure how to pass the pointer to fix this.
Thanks

yparitcher:
`font_info = NULL` only nulls the local copy not `otFonts.otRegular`
leads to double free when calling `free_ot_font()` more than once

Fixes #36

### DIFF
--- a/fbink.c
+++ b/fbink.c
@@ -2675,25 +2675,25 @@ int
 	// NOTE: We make sure we free any previous allocation first!
 	switch (style) {
 		case FNT_REGULAR:
-			if (free_ot_font(otFonts.otRegular) == EXIT_SUCCESS) {
+			if (free_ot_font(&otFonts.otRegular) == EXIT_SUCCESS) {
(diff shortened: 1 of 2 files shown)


NiLuJe:
Thanks!

PR #39, opened 2019-10-03
Makefile: add check for CHOST

jdek:
Generally, cross compilation toolchain prefixes include the '-' at the end. Don't want to change your setup, so just fallback to check a more standard CHOST cross compile if CROSS_TC is unset.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,12 @@ ifdef CROSS_TC
 	# NOTE: This relies on GCC plugins! Enforce AR & RANLIB to point to their real binary and not the GCC wrappers if your TC doesn't support that!
 	AR:=$(CROSS_TC)-gcc-ar
 	RANLIB:=$(CROSS_TC)-gcc-ranlib
+else ifdef CROSS_COMPILE
+	CC:=$(CROSS_COMPILE)cc
+	CXX:=$(CROSS_COMPILE)cxx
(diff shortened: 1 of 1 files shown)


geek1011:
I agree that CROSS_TC is less standard than CHOST (I've only seen it in NiLuJe's stuff), but the only place I've seen CHOST is in Gentoo stuff and a few projects. But, I've seen CROSS_PREFIX a bit more, and CROSS_COMPILE is used in the Linux kernel (with the dash).

jdek:
I'm not really bothered if you want to call it `CHOST`, `CROSS_PREFIX` or `CROSS_COMPILE`, just that it doesn't add an extra `-`. Which would you merge?

NiLuJe:
Yeah, `CHOST` is out because it has a very specific meaning in Gentoo, and you basically shouldn't ever set it yourself anywhere ;).

NiLuJe:
I have an extremely vague preference for `CROSS_PREFIX`, since it's also set in my own envs, but, since I'm usually using `CROSS_TC` myself, I have no stron opinion, both `CROSS_PREFIX` or `CROSS_COMPILE` would be fine with me ;).

jdek:
I set it as `CROSS_COMPILE` so it's similar to the kernel.

NiLuJe:
That's perfectly fine with me, thanks! ;).

Issue #40, opened 2019-10-13
Automated Go bindings & Errors


shermp:
Hi @NiLuJe 

I've started work on automated bindings using *c-for-go* here: https://github.com/shermp/FBInk/tree/go-bindings

I'm at the point where I seem to have a working binary, at least with the very limited testing I've done with it.

What I haven't decided on is how I'm actually going to deal with errors. Currently, the functions just return ints with no context. 

I hate to say it, but dealing with C `errno`'s is a bit of a PITA, doubly so when they're negative. To get the errno constants in c-for-go, one would have to find and include the appropriate `errno.h` file, which is compiler specific etc.

Have you any ideas on how I might be able to deal with FBInk errors, hopefully without too mich manual coding?

geek1011:
I did notice that almost every error has a WARN before it, so maybe that could be modified to save the warning into a global thread-local variable accessed by the Go wrapper, or something of that sort?

shermp:
Hmm, actually, the approach I took with my handwritten bindings should do the trick:
```
package fbink

/*
#cgo LDFLAGS: -L${SRCDIR}/../../Release -lfbink -lm
#include <stdlib.h>
#include <errno.h>
#include "../../fbink.h"
*/
import "C"

// Error codes as returned by FBInk
const (
	ExitSuccess int32 = int32(C.EXIT_SUCCESS)
	ExitFailure int32 = int32(C.EXIT_FAILURE) * -1
	ENoDev      int32 = int32(C.ENODEV) * -1
	ENotSup     int32 = int32(C.ENOTSUP) * -1
	ENoData     int32 = int32(C.ENODATA) * -1
	ETime       int32 = int32(C.ETIME) * -1
	EInval      int32 = int32(C.EINVAL) * -1
	EIlSeq      int32 = int32(C.EILSEQ) * -1
	ENoSpc      int32 = int32(C.ENOSPC) * -1
)
```
There should be minimal, if any error changes between FBInk releases, so this shouldn't be an issue.

geek1011:
> EDIT: This is going to be a very low level wrapper, so I'm not going to worry about wrapping errors in an actual error type or anything this time around.

Oh, I see what you're getting at. I was thinking you were wanting full error messages and such.

> 	ExitSuccess int32 = int32(C.EXIT_SUCCESS)

Yes, that should work. That's the approach I usually take when doing that kind of thing (but I make it a type and add an Error method to it for a nicer message).

shermp:
[...]

PR #41, opened 2020-01-23
[RFC] Remarkable support

tcrs:
I'm afraid there is much I don't understand about eink screens. I added a rather arbitrary mapping of the waveform modes in get_wfm_mode. I'm not sure what the dither bit should be set to in the update ioctl. Among other things...

Could you take a look and give me some advice on this? What I've got at the moment is happily displaying some text which is the most important hurdle for me at the moment (I'm porting this to use in my koreader port really).

---
This change is [Reviewable](https://reviewable.io/reviews/niluje/fbink/41)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,7 @@ ifdef FBINK_VERSION
 					LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION) for Kobo"'
 				else
 					ifdef REMARKABLE
-						LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION) for Remarkable"'
+						LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION) for reMarkable"'
 					else
 						# NOTE: Should never happen!
 						LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION)"'
--- a/Makefile
+++ b/Makefile
@@ -243,7 +243,7 @@ endif
 ifdef KOBO
 	TARGET_CPPFLAGS+=-DFBINK_FOR_KOBO
 endif
-# Toggle remarkable support
+# Toggle reMarkable support
 ifdef REMARKABLE
 	TARGET_CPPFLAGS+=-DFBINK_FOR_REMARKABLE
 endif
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,9 @@ ifndef LINUX
 ifndef CERVANTES
 ifndef LEGACY
 ifndef KINDLE
+ifndef REMARKABLE
 	KOBO=true
+endif
 endif
 endif
 endif
@@ -241,6 +243,10 @@ endif
 ifdef KOBO
 	TARGET_CPPFLAGS+=-DFBINK_FOR_KOBO
 endif
+# Toggle remarkable support
+ifdef REMARKABLE
+	TARGET_CPPFLAGS+=-DFBINK_FOR_REMARKABLE
+endif
 
 # And that should definitely be honored by everything, so, add it to EXTRA_CPPFLAGS
 EXTRA_CPPFLAGS+=$(TARGET_CPPFLAGS)
@@ -262,8 +268,12 @@ ifdef FBINK_VERSION
 				ifdef KOBO
 					LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION) for Kobo"'
 				else
-					# NOTE: Should never happen!
-					LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION)"'
+					ifdef REMARKABLE
+						LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION) for Remarkable"'
+					else
+						# NOTE: Should never happen!
+						LIB_CFLAGS+=-DFBINK_VERSION='"$(FBINK_VERSION)"'
+					endif
 				endif
 			endif
 		endif
@@ -553,6 +563,9 @@ cervantes:
 	$(MAKE) strip CERVANTES=true
 	$(CURDIR)/tools/do_debian_package.sh $(OUT_DIR) armel
 
+remarkable:
+	$(MAKE) strip REMARKABLE=true
+
 libunibreak.built:
 	mkdir -p LibUniBreakBuild
 	cd libunibreak && \
@@ -667,4 +680,4 @@ distclean: clean libunibreakclean
 	rm -rf LibUniBreakBuild
 	rm -rf libunibreak.built
 
-.PHONY: default outdir all staticlib sharedlib static shared striplib striparchive stripbin strip debug static pic shared release kindle legacy cervantes linux armcheck kobo libunibreakclean utils alt dump clean distclean
+.PHONY: default outdir all staticlib sharedlib static shared striplib striparchive stripbin strip debug static pic shared release kindle legacy cervantes linux armcheck kobo remarkable libunibreakclean utils alt dump clean distclean
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This is intended to fill the void felt by Kobo developers and tinkerers when the
 It's especially cruel when moving to a Kobo, after being used to the ubiquity of `eips` on Kindle...
 
 In short, it prints messages or images on your screen, handling the low-level tinkering with both the Linux framebuffer interface, and the i.MX EPDC.  
-It's been tested on Kobo, Kindle and BQ Cervantes, but porting it to other Linux, i.MX eInk devices should be trivial (hell, even Sipix support shouldn't be too hard).
+It's been tested on Kobo, Kindle, BQ Cervantes and reMarkable, but porting it to other Linux, i.MX eInk devices should be trivial (hell, even Sipix support shouldn't be too hard).
 
 By default, text rendering relies on bundled fixed cell bitmap fonts ([see this post](https://www.mobileread.com/forums/showpost.php?p=3765426&postcount=31) for a small sampling),
 but thanks to [@shermp](https://github.com/shermp)'s contributions ([#20](https://github.com/NiLuJe/FBInk/pull/20)), you can also rely on full-fledged TrueType/OpenType font rendering!
@@ -88,6 +88,7 @@ The choice of target platform is handled via a simple variable:
 -   Pass `KINDLE=1` to make for a Kindle build (`make kindle` does that on a stripped static build).
 -   Pass `KINDLE=1 LEGACY=1` to make for a FW 2.x Kindle build (`make legacy` does that on a stripped static build). This basically just disables CLOEXEC, which might not be supported on FW 2.x.
 -   Pass `CERVANTES=1` to make for a BQ/Cervantes build (`make cervantes` does that on a stripped static build).
+-   Pass `REMARKABLE=1` to make for a reMarkable build (`make remarkable` does that on a stripped static build).
 
 The same logic is used to allow for a bit of tailoring:
 -   Pass `MINIMAL=1` to make for a build with limited functionality (only fixed cell font rendering, no image rendering, no extra fonts, no OpenType), which yields a much smaller application & library.
@@ -113,6 +114,8 @@ Kindle support covers the full Kindle lineup, starting from the K2.
 Kobo support covers the full Kobo lineup, starting from the Kobo Touch A/B/C.
 
 BQ Cervantes support has been contributed by [@pazos](https://github.com/pazos) ([#17](https://github.com/NiLuJe/FBInk/pull/17)), and should handle the current lineup.
+
+reMarkable support has been contributed by [@tcrs](https://github.com/tcrs) ([#41](https://github.com/NiLuJe/FBInk/pull/41)).
 
 ## Bindings in other languages
 
--- a/eink/mxcfb-remarkable.h
+++ b/eink/mxcfb-remarkable.h
@@ -20,8 +20,8 @@
 
 /* Original source:
  * https://github.com/reMarkable/linux/blob/zero-gravitas/include/uapi/linux/mxcfb.h
- * added waveform modes from libremarkable:
- * https://github.com/canselcik/libremarkable/blob/master/src/framebuffer/common.rs#L338
+ * Waveform modes constants from libremarkable:
+ * https://github.com/canselcik/libremarkable
  */
 
 /*
@@ -100,23 +100,31 @@ struct mxcfb_rect {
 #define UPDATE_MODE_PARTIAL			0x0
 #define UPDATE_MODE_FULL			0x1
 
-// These modes added by copying values from libremarkable
+// Findings courtesy of libremarkable
+// c.f., https://github.com/canselcik/libremarkable/blob/67ff7ea3926319a6d33a216a2b8c1f679916aa3c/src/framebuffer/common.rs#L338
 #define WAVEFORM_MODE_INIT			0
 #define WAVEFORM_MODE_DU			1
 #define WAVEFORM_MODE_GC16			2
-#define WAVEFORM_MODE_GC16_FAST		3
-#define WAVEFORM_MODE_GL16_FAST		6
+#define WAVEFORM_MODE_GC16_FAST			3
+
+#define WAVEFORM_MODE_GL16_FAST			6
 #define WAVEFORM_MODE_DU4			7
 #define WAVEFORM_MODE_REAGL			8
-#define WAVEFORM_MODE_REAGLD		9
+#define WAVEFORM_MODE_REAGLD			9
 #define WAVEFORM_MODE_GL4			0xA
-#define WAVEFORM_MODE_GL16_INV		0xB
+#define WAVEFORM_MODE_GL16_INV			0xB
 
+// NOTE: This might actually be A2?
+//       On every other platform, GLR16 == REAGL && GLD16 == REAGLD
 #define WAVEFORM_MODE_GLR16			4
+// NOTE: Which would actually make this GL16?
 #define WAVEFORM_MODE_GLD16			5
 #define WAVEFORM_MODE_AUTO			257
 
 #define TEMP_USE_AMBIENT			0x1000
+
+// Again, pilfered from libremarkable ;).
+#define TEMP_USE_REMARKABLE			0x0018
 
 #define EPDC_FLAG_ENABLE_INVERSION		0x01
 #define EPDC_FLAG_FORCE_MONOCHROME		0x02
--- a/eink/mxcfb-remarkable.h
+++ b/eink/mxcfb-remarkable.h
@@ -1,0 +1,211 @@
+/*
+ * Copyright (C) 2013-2015 Freescale Semiconductor, Inc. All Rights Reserved
+ */
+
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+/*
+ * @file uapi/linux/mxcfb.h
+ *
+ * @brief Global header file for the MXC frame buffer
+ *
+ * @ingroup Framebuffer
+ */
+#ifndef __ASM_ARCH_MXCFB_H__
+#define __ASM_ARCH_MXCFB_H__
+
+#include <linux/fb.h>
+
+#define FB_SYNC_OE_LOW_ACT	0x80000000
+#define FB_SYNC_CLK_LAT_FALL	0x40000000
+#define FB_SYNC_DATA_INVERT	0x20000000
+#define FB_SYNC_CLK_IDLE_EN	0x10000000
+#define FB_SYNC_SHARP_MODE	0x08000000
+#define FB_SYNC_SWAP_RGB	0x04000000
+#define FB_ACCEL_TRIPLE_FLAG	0x00000000
+#define FB_ACCEL_DOUBLE_FLAG	0x00000001
+
+struct mxcfb_gbl_alpha {
+	int enable;
+	int alpha;
+};
+
+struct mxcfb_loc_alpha {
+	int enable;
+	int alpha_in_pixel;
+	unsigned long alpha_phy_addr0;
+	unsigned long alpha_phy_addr1;
+};
+
+struct mxcfb_color_key {
+	int enable;
+	__u32 color_key;
+};
+
+struct mxcfb_pos {
+	__u16 x;
+	__u16 y;
+};
+
+struct mxcfb_gamma {
+	int enable;
+	int constk[16];
+	int slopek[16];
+};
+
+struct mxcfb_gpu_split_fmt {
+	struct fb_var_screeninfo var;
+	unsigned long offset;
+};
+
+struct mxcfb_rect {
+	__u32 top;
+	__u32 left;
+	__u32 width;
+	__u32 height;
+};
+
+#define GRAYSCALE_8BIT				0x1
+#define GRAYSCALE_8BIT_INVERTED			0x2
+#define GRAYSCALE_4BIT                          0x3
+#define GRAYSCALE_4BIT_INVERTED                 0x4
+
+#define AUTO_UPDATE_MODE_REGION_MODE		0
+#define AUTO_UPDATE_MODE_AUTOMATIC_MODE		1
+
+#define UPDATE_SCHEME_SNAPSHOT			0
+#define UPDATE_SCHEME_QUEUE			1
+#define UPDATE_SCHEME_QUEUE_AND_MERGE		2
+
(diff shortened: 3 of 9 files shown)


NiLuJe:
Looks fairly okay for a first stab ;).

* I haven't double-checked the mxcfb header, but we don't actually need much of it to be "right" ;).
You're possibly missing a few waveform mode constants (c.f., https://github.com/canselcik/libremarkable/blob/master/src/framebuffer/common.rs#L338).
Switching to the "better" TEMP flag might be nice, too.

* Speaking of waveform mode constants, you should be able to use a 1:1 mapping in get_wfm_mode (the enum should have all the required names already, between the Kindle & Kobo support ;)).
Anything that's in the header should be made available here (even the dubiously useful ones, like INIT ^^), you don't have to specifically care whether they currently do something useful on the device.

* Related note: I'm slightly confused by the libremarkable notes around GLR16/GLD16. In all the devices I'm familiar with, GLR16 == REAGL and GLD16 == REAGLD. A2 is something else entirely (very specifically a 2-bit mode: black or white, and nothing else. That downside allows it to be the fastest refresh mode, even when you can throw dithering in the mix).

* Assuming it behaves properly (which isn't a given, it apparently doesn't on *some* Kindles), look at the mk7 refresh call for the intricacies related to dithering/quant_bit ;).

You only care about the "v2" dithering modes (in libremarkable, PASSTHROUGH & DRAWING). I don't actually know what to make of "DRAWING" though. On my usual devices; that'd be Floyd-Steinberg, but I've never seen anything else than ORDERED being supported (there's a dmesg warning when you try to use an unimplemented dithering mode).

NiLuJe:
At a very quick glance at the driver, the actual dithering value doesn't matter much?

https://github.com/reMarkable/linux/blob/zero-gravitas/drivers/video/fbdev/mxc/mxc_epdc_v2_fb.c#L2605

https://github.com/reMarkable/linux/blob/zero-gravitas/include/uapi/linux/mxcfb.h

(I'm not on my main dev box, and Kindle/Kobo kernels are only available as tarballs, so I can't have a quick comparative look at them, so take this with a grain of salt ^^).

NiLuJe:
It's also highly unclear to me which version of the driver it's actually supposed to use, especially when looking at the commits in the zero-gravitas_4.9 branch, which mostly appear to target the older driver (https://github.com/reMarkable/linux/blob/lars/zero-gravitas_4.9/drivers/video/fbdev/mxc/mxc_epdc_fb.c && https://github.com/reMarkable/linux/blob/lars/zero-gravitas_4.9/include/uapi/linux/mxcfb.h).

In which case, that dithering flag is actually meaningless, and I have no idea why the kernel actually support the ioctl with that update,d larger struct.

(Although I have seen some weird hoop-jumping being done before in the name of backward compatibility, so, who knows ^^).

NiLuJe:
The [config](https://github.com/reMarkable/linux/blob/zero-gravitas/arch/arm/configs/zero-gravitas_defconfig) appear to confirm that the v2 driver is unused, so you can pretty much forget about, all that, and always set dithering & quant_bit to 0.

Still puzzled as to why they kept the updated struct, though.

But I can roll with that, it's certainly less messy than what I've seen before, which is either different ioctl for each device, or a device with support for 2 or 3 different ioctls with a slightly different struct each time ;).

NiLuJe:
I wouldn't mind a link to the upstream Kernel repo in the bundled mxcfb copy, by the way ;).

I probably ought to do that myself for the Kindle & Kobo ones, actually ^^.

NiLuJe:
Extremely minor nitpick: I'm guessing they care about the peculiar casing of the `reMarkable` name, so I'd double-check that, too ;).

NiLuJe:
And, obviously, because I kind of forgot to say it in my excitement: thanks for working on this ;).

NiLuJe:
I'd also add a small `identify_remarkable` to fbink_id.c, just to set the name/codename (which appears to be `zero-gravitas`), and to set the DPI (226?).

It'll also come in handy if you want to implement fbdepth (as being able to switch to 8bpp would improve performance in KOReader, because RGB565 is the worst).
To that end, I wouldn't mind having a peek at the console output from a random print, mainly to check what the "standard" rotation is ;).

tcrs:
Thanks for the feedback, I've had a go at the improvements you suggested. The waveform mode mapping is still not 1:1.

Here's the output from a simple fbink invocation, this is just after starting xochitl so should be set up in the default way.

```
remarkable: ~/ ./fbink 'HELLLO'
[FBInk] Detected a reMarkable
[FBInk] Clock tick frequency appears to be 100 Hz
[FBInk] Screen density set to 226 dpi
[FBInk] Variable fb info: 1404x1872, 16bpp @ rotation: 1 (Clockwise, 90°)
[FBInk] Fontsize set to 24x24 (IBM base glyph size: 8x8)
[FBInk] Line length: 58 cols, Page size: 78 rows
[FBInk] Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes
[FBInk] Pen colors set to #000000 for the foreground and #FFFFFF for the background
Printing string 'HELLLO' @ column 0 + 0px, row 0 + 0px (overlay: N, no BG: N, no FG: N, inverted: N, flashing: N, centered: N, halfway: N, left padded: N, right padded: N, clear screen: N, waveform: AUTO, dithered: N, nightmode: N, skip refresh: N, font: 0, font scaling: x0)
Next line should start @ row 1
```

tcrs:
Also, I'm using `fbdepth` when launching `koreader` now to set the panel to 8bpp and it seems to be working, does appear a bit snappier (but I could be imagining it...)

NiLuJe:
Great, thanks :).

Don't worry too much about the waveform mapping, I'll take a pass at it tomorrow if I can, that'll probably be easier than a few more back & forth, especially since you don't have any other device to compare to ;).
I'll just ask you to double-check that I didn't do anything obviously stupid when I'm done ;).

----

Good to know about fbdepth (and it being snappier is definitely the end goal. A quick, obvious test is panning a full-screen cover, for instance. You also get free software dithering as a bonus ;).).
Quick question, though: how are you calling it, exactly?
I assume you're omitting the -r, --rota switch?

I'll probably also set a few flags to make that not entirely broken, to that end, you can try running the "rota" tool from the utils folder (it should be build alongside fbdepth w/ the utils target), that should confirm whether there's any weird crap happening on rotation like on some Kobo devices ;).

tcrs:
I'm currently just using fbdepth to set the depth yea, I've so far left the rotation at the default.

Here's the output of rota, I assume it means more to you that it does to me :)
```
remarkable: ~/ ./rota 
Variable fb info: 1404x1872, 16bpp @ rotation: 1 (Clockwise, 90°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes

FB_ROTATE_UR to FB_ROTATE_CCW, +1 increments
Setting rotate to 0 (Upright, 0°)
Rotate is now 0 (Upright, 0°)
Variable fb info: 1872x1404, 16bpp @ rotation: 0 (Upright, 0°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 3776 bytes
Setting rotate to 1 (Clockwise, 90°)
Rotate is now 1 (Clockwise, 90°)
Variable fb info: 1404x1872, 16bpp @ rotation: 1 (Clockwise, 90°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes
Setting rotate to 2 (Upside Down, 180°)
Rotate is now 2 (Upside Down, 180°)
Variable fb info: 1872x1404, 16bpp @ rotation: 2 (Upside Down, 180°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 3776 bytes
Setting rotate to 3 (Counter Clockwise, 270°)
Rotate is now 3 (Counter Clockwise, 270°)
Variable fb info: 1404x1872, 16bpp @ rotation: 3 (Counter Clockwise, 270°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes

FB_ROTATE_UR to FB_ROTATE_CCW, +2 increments
Setting rotate to 0 (Upright, 0°)
Rotate is now 0 (Upright, 0°)
Variable fb info: 1872x1404, 16bpp @ rotation: 0 (Upright, 0°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 3776 bytes
Setting rotate to 2 (Upside Down, 180°)
Rotate is now 2 (Upside Down, 180°)
Variable fb info: 1872x1404, 16bpp @ rotation: 2 (Upside Down, 180°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 3776 bytes

FB_ROTATE_CW to FB_ROTATE_CCW, +2 increments
Setting rotate to 1 (Clockwise, 90°)
Rotate is now 1 (Clockwise, 90°)
Variable fb info: 1404x1872, 16bpp @ rotation: 1 (Clockwise, 90°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes
Setting rotate to 3 (Counter Clockwise, 270°)
Rotate is now 3 (Counter Clockwise, 270°)
Variable fb info: 1404x1872, 16bpp @ rotation: 3 (Counter Clockwise, 270°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes

FB_ROTATE_UR to FB_ROTATE_CCW, +2 increments, intermerdiary rota if ==
Setting rotate to 0 (Upright, 0°)
Rotate is now 0 (Upright, 0°)
Variable fb info: 1872x1404, 16bpp @ rotation: 0 (Upright, 0°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 3776 bytes
Setting rotate to 2 (Upside Down, 180°)
Rotate is now 2 (Upside Down, 180°)
Variable fb info: 1872x1404, 16bpp @ rotation: 2 (Upside Down, 180°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 3776 bytes

FB_ROTATE_CW to FB_ROTATE_CCW, +2 increments, intermerdiary rota if ==
Setting rotate to 1 (Clockwise, 90°)
Rotate is now 1 (Clockwise, 90°)
Variable fb info: 1404x1872, 16bpp @ rotation: 1 (Clockwise, 90°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes
Setting rotate to 3 (Counter Clockwise, 270°)
Rotate is now 3 (Counter Clockwise, 270°)
Variable fb info: 1404x1872, 16bpp @ rotation: 3 (Counter Clockwise, 270°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 10813440 bytes & line length: 2816 bytes
```

Let me know if there is any other things it would be useful to see the output of.

tcrs:
I just added some default rotation settings to the deviceQuirks which seem to work OK. I'm not sure if I should be setting `ntxRotaQuirk = NTX_ROTA_SANE` as the comment says that expects boot rotation to be 0 I think. These settings make `fbdepth -r -1` set the rotation to the proper portrait rotation for the remarkable at least.

NiLuJe:
That was precisely what I was planning to do ;).

NiLuJe:
Although, yeah, the comment in the enum is a bit misleading, as it's specific to the Kobo Libra.

In practice, it means no rotation shenanigans, and honor bootRota as the "native" Portrait.

----

I'll add a bunch of comments to make that clearer ;).

NiLuJe:
Okay, I've *hopefully* untangled the `get_wfm_mode` mapping...

As a quick test, because it might come in handy for KOReader to have a definitive answer on that front, can you check if displaying an image w/ A2 (i.e., `fbink -W A2 -i img.png`) actually crushes it to a bitonal result?

tcrs:
> Okay, I've _hopefully_ untangled the `get_wfm_mode` mapping...

Thanks :)

> As a quick test, because it might come in handy for KOReader to have a definitive answer on that front, can you check if displaying an image w/ A2 (i.e., `fbink -W A2 -i img.png`) actually crushes it to a bitonal result?

Yea it looks like it does, drawing with A2 on a cleared screen just draws the black bits in the image. Running again with the default settings draws full details.

NiLuJe:
@tcrs: Yay!

That'll allow me to play with the constant names a bit in the header to make this all slightly less confusing, and this is good to go, thanks :).

I'll be tagging a release soon-ish, but if that doesn't happen soon enough for the KOReader bits, don't hesitate to bump FBInk to a post-merge commit in koreader-base ;).

NiLuJe:
As an additional test, does something similar happens with `GL16`?

(I'm wondering if it's actually GL16, or GC4).

In either case, it's a far less interesting mode in practice, and it can be a bit weird, so do try after a *separate* screen clear (because if the fb content doesn't "change", both of 'em are optimized away, so a double-refresh with those is essentially a no-op, even if the content was actually redrawn, as long as it's identical).

tcrs:
> As an additional test, does something similar happens with `GL16`, or do you get proper shades of gray?
> 
> (I'm wondering if it's actually GL16, or GC4).
> 
> In either case, it's a-W GL16 -i fbink_image.png far less interesting mode in practice (well, GL16 has its use in theory, but not in practice on a device with GL16_FAST), and it can be a bit weird, so do try after a _separate_ screen clear (because if the fb content doesn't "change", both of 'em are optimized away, so a double-refresh with those is essentially a no-op, even if the content was actually redrawn, as long as it's identical).

`./fbink -c; ./fbink -W GL16 -i fbink_image.png` draws the image but it flickers constantly for about 10 seconds then settles into a poor reproduction of the image but with a few gray shades. Looks pretty worrying when it is happening! `GL16_FAST` looks basically the same to my eyes (lots of flickering followed by a bad image).

I saw a similar effect when I was first porting koreader (and just choosing the waveform modes at random).

NiLuJe:
Huh. That's... weird :D. But good to know nonetheless, thanks ;).

NiLuJe:
Sidebar: if you (or anyone else with shell access to a reMarkable, really) want to dig into this further, I can whip up the usual strace patch to decode mxcfb ioctls, and then it's just a matter of strace'ing the official app while doing various different things to see what exactly it's using ;).

NiLuJe:
Okay, said [strace patch](http://trac.ak-team.com/trac/browser/niluje/Configs/trunk/Kindle/Misc/strace-mxcfb-ioctls-remarkable.patch) is done ;).

It should then be as easy as running `./strace -fitv -e trace=ioctl -X verbose -p $(pidof xochitl)` to see what makes it tick ;).

[strace-remarkable.tar.gz](https://github.com/NiLuJe/FBInk/files/4119913/strace-remarkable.tar.gz)

NiLuJe:
Hmm, see also https://github.com/NiLuJe/FBInk/commit/ffcd9830f2a8c653b99b061c8dd8c78a3ad3a384

Fair warning: I'm terrible at this stuff, and I don't have an IDA Pro license, so this is cobbled together via objdump ;).

Speaking of, @tcrs, I wouldn't mind an updated copy of that (libqsgepaper.a & xochitl), as what's available in the [libremarbale repo](https://github.com/canselcik/libremarkable/tree/master/reference-material) is over two years old ;).

tcrs:
Here's a few captures using that strace binary & the options you said. I grepped all the mxcfb ioctls out of the full trace (there's constant trace noise from some networking stuff going on).

This one is me just moving around the ui, opening settings, searching etc.
[ui.log](https://github.com/NiLuJe/FBInk/files/4124278/ui.log)

This is drawing a load of random squiggles and then erasing most of them again
[drawing.log](https://github.com/NiLuJe/FBInk/files/4124281/drawing.log)

This is rendering a single page of a pdf (xochitl really does take about 30s to do that)...
[pdf.log](https://github.com/NiLuJe/FBInk/files/4124283/pdf.log)

NiLuJe:
Oh, cool, thanks ;).

That seems to confirm what I dug out of `libqsgepaper.a`, and that the dither_mode/quant_bit fields are uninitialized here, which leads to a lot of confusing garbage if you try to make sense of it ;p.

The only question mark left being that "Highlight" mode (currently mapped to REAGL), which I'd take to mean A2?
How does it fare with image content w/ FBInk (same test as before ;)).

NiLuJe:
Oh, err, it turns out the actual header is part of the SDK... \o/

```C
#ifndef EPFRAMEBUFFER_H
#define EPFRAMEBUFFER_H

#include <QtCore/QElapsedTimer>
#include <QtCore/QFile>
#include <QtCore/QMutex>
#include <QtCore/QObject>
#include <QtGui/QImage>

#include <atomic>
#include <mutex>

class EPFrameBuffer : public QObject
{
    Q_OBJECT

    // Testing results:
    // 0: ~780ms, initialize to white
    // 1: ~170ms, mono, no flashing
    // 2: ~460ms, 15 grayscale, flashing all pixels white -> black pixels
    // 3: ~460ms, 15, grayscale, flashing all pixels white -> black pixels
    // 4: ~460ms, 4 grayscale, flashing all pixels white -> black pixels
    // 5: ~460ms, 4 grayscale, flashing all pixels white -> black pixels
    // 6: ~135ms, no flashing
    // 7: ~300ms, white, 2 gray, 1 black, no flashing, stuck black?
    // 8: ~435ms, fast flashing, all gray, highlight, several grays, lowest color - 1 gets stuck
    // 9: ~2365ms, lots of flashing, delta something?

    enum Waveform {
        /** Display initialization
          * All -> white
          * 2000ms
          * ~780ms stall in driver
          * Supposedly low ghosting, high ghosting in practice.
          */
        INIT = 0,

        /**
          * Monochrome menu, text input, pen input
          * All -> black/white
          * 260ms to complete
          * 170ms stall in driver
          * Low ghosting
          */
        DU = 1,

        /** High quality images
          * All -> all
          * 480ms to complete
          * 460ms stall in driver
          * Very low ghosting
          */
        GC16 = 2,

        /** Text with white background
          * 16 -> 16
          * 480ms to complete
          * 460ms stall in driver
          * Medium ghosting
          *
          * Local update when white to white, Global update when 16 gray levels
          * For drawing anti-aliased text with reduced flash. This waveform makes full screen update
          * - the entire display except pixels staying in white will update as the new image is
          * written.
          */
        GL16 = 3,

        /** Text with white background
          * All -> all
          * 480ms to complete
          * 460ms stall in driver
          * Low ghosting
          *
          * Global Update
          * For drawing anti-aliased text with reduced flash and image artifacts (used in
          * conjunction with an image preprocessing algorithm, very little ghosting). Drawing time
          * is around 600ms. This waveform makes full screen update - all the pixels are updated or
          * cleared. Use this for anti-aliased text.
          */
        GLR16 = 4,

        /** Text and graphics with white background
          * All -> all
          * 480ms to complete
          * 460ms stall in driver
          * Low ghosting
          *
          * Global Update
          * For drawing anti-aliased text with reduced flash and image artifacts (used in
          * conjunction with an image preprocessing algorithm, very little ghosting) even more
          * compared to the GLR16 mode. Drawing time is around 600ms. This waveform makes full
          * screen update - all the pixels are updated or cleared. Use this for anti-aliased text.
          */
        GLD16 = 5,

        /** Fast page flipping at reduced contrast
          * [0, 29, 30, 31] -> [0, 30]
          * 120ms to complete
          * 135ms stall in driver
          * Medium ghosting
          *
          * Local Update
          * For drawing black and white 2-color images. This waveform supports transitions from and
          * to black or white only. It cannot be used to update to any graytone other than black or
          * white. Drawing time is around 120ms. This waveform makes partial screen update - it
          * updates only changed pixels. Image quality is reduced in exchange for the quicker
          * response time. This waveform can be used for fast updates and simple animation.
          */
        A2 = 6,

        /** Anti-aliased text in menus, touch and pen input
          * All -> [0, 10, 20, 30]
          * 290ms to complete
          * 300ms stall in driver
          *
          * No flashing, black seems to be stuck afterwards?
          */
        DU4 = 7,

        /** Unknown1
          * 435ms stall in driver
          *
          * Fast flashing, all gray, highlight?
          * Multiple grays
          * Next to lowest color gets stuck
          */
        UNKNOWN = 8,

        /** Init?
          * 2365ms stall in driver
          *
          * Lots of flashing, seems to do some delta updates (inverting unchanged?)
          */
        INIT2 = 9
    };

public:
    static EPFrameBuffer *instance();

    /// Which waveforms we use for different kinds of updates
    enum WaveformMode {
        Initialize = INIT,
        Mono = DU,
        Grayscale = GL16,
        HighQualityGrayscale = GC16,
        Highlight = UNKNOWN
    };

    enum UpdateMode {
        PartialUpdate = 0x0,
        FullUpdate = 0x1
    };

    static QImage *framebuffer() {
        return &instance()->m_fb;
    }

    Q_INVOKABLE static void setForceFull(bool force) { instance()->m_forceFull = force; }
    static bool isForceFull() { return instance()->m_forceFull; }

    int lastUpdateId() const { return m_lastUpdateId; }

    void setSuspended(bool suspended) { m_suspended = suspended; }
    bool isSuspended() const {
        std::lock_guard<std::mutex> locker(fbMutex);
        return m_suspended;
    }

    mutable std::mutex fbMutex;

    qint64 timeSinceLastUpdate() const;

public slots:
    static void clearScreen();
    static void sendUpdate(QRect rect, WaveformMode waveform, UpdateMode mode, bool sync = false);
    static void waitForLastUpdate();

private:
    EPFrameBuffer();
    QImage m_fb;
    QFile m_deviceFile;
    bool m_forceFull = false;
    bool m_suspended = false;
    int m_lastUpdateId = 0;

    std::mutex m_timerLock;
    QElapsedTimer m_lastUpdateTimer;
};

#endif // EPFRAMEBUFFER_H
```

NiLuJe:
Okay, so I've updated the mappings according to that...

Which means we're unfortunately back to square one as far as A2 is concerned...

So, mildly intrigued how `fbink -cfw && fbink -W A2 -i img.png` fares (compared mainly to DU & REAGL).

On the vague chance the REAGL ones behave like on Kindle, you may need to make 'em flashing/FULL (-f) for them to behave properly. In which case, they won't behave very nicely when refreshing identical content...

tcrs:
> Okay, so I've updated the mappings according to that...
> 
> Which means we're unfortunately back to square one as far as A2 is concerned...
> 
> So, mildly intrigued how `fbink -cfw && fbink -W A2 -i img.png` fares (compared mainly to `DU` & `REAGL`).
> 
> On the vague chance the REAGL ones behave like on Kindle, you may need to make 'em flashing/FULL (`-f`) for them to behave properly. In which case, they won't behave very nicely when refreshing identical content, hence the full clear beforehand ;)...

I built fbink from master and tried `./fbink -cfw && ./fbink -W A2 -i fbink_image.png` and `./fbink -cfw && ./fbink -W A2 -f -i fbink_image.png`. Both do ~10s of disturbing flickering of the update area and then settle down to a (different but pretty poor) representation of the image (but using multiple gray shades). REAGL/REAGLD behave the same as A2 but with different image reproduction after settling (REGL seems to end up more or less inverted).

So no idea what that means!

NiLuJe:
Huh. For reference, if I take my usual [eInk dithering torture test](https://github.com/koreader/koreader-base/pull/800#issuecomment-459877277), except in Grayscale:

Forma_Gray

That's roughly how I expect it to get rendered with DU or A2:

Forma_DU

----

Soo, I'm guessing a full recap is in order, with a test of every waveform mode, in order: `DU`, `GC16`, `GL16` (which I expect to work properly, ass far a crazy flashing is concerned ^^), and then `REAGL`, `REAGLD`, `A2`, `DU4`, `UNKNOWN`, `INIT2`.

I'm mainly concerned with identifying where the "actual" A2 behavior is hiding (if any).

----

FWIW, GC16 & GL16 should render something like the [first picture here](https://github.com/koreader/koreader/pull/4541#issuecomment-459955126) (i.e., full of banding).

----

CC @canselcik, because most of what was said here might come in handy for libremarkable in one shape or form (also, more guinea pigs :D).

tcrs:
OK, something funny is going on! I tested all the modes with
```
./fbink -fk && ./fbink -W MODE -i img.jpg
```
I took a load of pictures, which ended up with `DU` & `REAGL` looking like the example `DU`/`A2` image you posted, `GC16` & `GL16` ended up banded grayscale. All the other modes did terrible flickering for 10s and settled into a banded randomly inverted image.

Then I thought I'd double check and ran the tests again. Now they *all* produce something which looks like `GC16` without any crazy flickering. Am I doing something stupid in my testing here?

NiLuJe:
I've sometimes seen weird things happening w/ A2 , but not to that extent ;).

That said, I'd try with the `-w` flag (on both sides): this ensures FBInk will block until the driver has *completely* finished dealing with the refresh, which *should* prevent the driver from deciding to merge a couple of refreshes together, which would ultimately affect the effective waveform selection ;).

FWIW, `REAGL` behaving like `DU` is consistent with the very first test you did when I originally asked about A2. (It also vaguely makes sense as it's the next idea after the properly identified mode (i.e., those used by xochitl).

tcrs:
Ah yes the `-w` option on the clear makes things consistent and match my first run. I think on the first run I was running the clear as a separate command (rather than linking them with `&&`) so there would have been a longer pause between the two. Makes sense now, mystery solved thanks!

Using `./fbink -fk -w && ./fbink -w -W MODE -i eink_test.png` I get the results: `DU `& `REAGL` look like the example `DU`/`A2` image you posted, `GC16` & `GL16` end up banded grayscale like the image you pointed to.

All the other modes have terrible flickering for 10s and settle into a banded randomly inverted image like this:
flicker_modes
and also timeout waiting for the update to complete.

NiLuJe:
Whew, mystery solved ;).

Many thanks for the tests!

I'll tweak the mapping for this A2 trickery and update the documentation to strongly discourage anyone from trying other values, then, thanks!

Because, ouch, that's some weird output there :D.

hdhbdh:
This fixes #40.

PR #51, opened 2021-04-19
Update README.md link to CMD options

fulldecent:
This change is [Reviewable](https://reviewable.io/reviews/niluje/fbink/51)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ See also the various bindings in other languages, which often include a few exam
 
 A CLI utility is available, built around the same public API that can be used via a shared or static library for C projects, or via FFI in other languages (beware, though, it's licensed under the GPLv3+, not the LGPL).  
 The gist of the documentation is composed of the various comments in the [public header](fbink.h), detailing how the API can be used, and what for. Don't hesitate to contact me if things appear unclear!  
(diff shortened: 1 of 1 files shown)


NiLuJe:
Reading unprocessed printf tokens requires a weird kind of dedication, but, why not ;o).

fulldecent:
That's right!

Separately, I would be happy to contribute a translation of the `printf` documentation into a command line readme in Markdown.

It didn't seem to fit in the main readme, but if you'd like it just tell me where to put it.

NiLuJe:
To be honest, I originally purposefully did NOT do that because my experience with that approach in [KindleTool](https://github.com/NiLuJe/KindleTool) lead to pain & misery with having to maintain three slightly different copies of the same thing (printf, md, man) ^^.

That said, given that I've almost run our of free getopt token in FBInk anyways, it's unlikely to see any massive changes, so, sure, go ahead ;).

(Because, to be fair, KindleTool happens to require adding stupid device ids to the help listing every other day, which kinda compounds the annoyance factor ^^).

NiLuJe:
As to where to put it, that's a good question. I'd say a CLI.md in the root folder, with a link to it in the README?

PR #59, opened 2021-04-26
Command line documentation

fulldecent:
# Live preview

https://github.com/fulldecent/FBInk/blob/patch-7/API.md

# Work plan:

- [x] Translate all command documentation from [C code](https://github.com/NiLuJe/FBInk/blob/master/fbink_cmd.c) to Markdown
  - Progress: everything is touched except the IMAGE section and below (i.e. https://github.com/NiLuJe/FBInk/blob/master/fbink_cmd.c#L281-L356)
- [x] Address every XXX and TODO on the page
- [ ] Consider if some options should be extracted from the "Options affecting the message's position on screen", "Options affecting the message's appearance" where they also apply to images into a more general section name.
- [ ] The synopses uses a code block inside a list. GitHub renders this poorly and/or incorrectly. Decide if something needs to be done about that.
- [ ] Reconcile differences between this new doc and existing documentation (help requested)

---

# Differences:

I have taken liberty in the interest of readability of reorganizing slightly the documentation. Every change, other than minor formatting is listed below. I recommend that these all be adopted as updates to the printf version (help requested with that).

- [ ] Additional synopses added. Certain useful modes don't use STRING, those are separated out into separate synopses lines.
- [ ] `--refresh` and `--cls` were moved to be adjacent and added under a title "Option for only clearing & refreshing screen".
- [ ] Moved OpenType & TrueType notes in as a new section under options.
- [ ] Moved Notes about multiple string arguments to after the options.

Every other bite-sized difference was submitted as a separate PR already.

---
This change is [Reviewable](https://reviewable.io/reviews/niluje/fbink/59)

### DIFF
--- a/API.md
+++ b/API.md
@@ -15,10 +15,10 @@
   Print image on your device's screen.
 
 * ```sh
-  fbink [-WHf] [-k] --refresh [top=NUM,left=NUM,width=NUM,height=NUM]
+  fbink [-WHf] --refresh [top=NUM,left=NUM,width=NUM,height=NUM]
   ```
 
-  Refresh the screen (or a region of it).
+  Refresh the screen as per your specification, without touching the framebuffer.
 
 * ```sh
   fbink [-BhHWDbw] --cls [top=NUM,left=NUM,width=NUM,height=NUM]
--- a/API.md
+++ b/API.md
@@ -354,25 +354,29 @@
 
   Honors `-h`, `--invert`; `-f`, `--flash`; `-c`, `--clear`; `-W`, `--waveform`; `-D`, `--dither`; `-H`, `--nightmode`; `-b`, `--norefresh`; `-m`, `--centered`; `-M`, `--halfway`; `-o`, `--overlay`; `-T`, `--fgless`; `-O`, `--bgless`; `-C`, `--color`; `-B`, `--background`; `-l`, `--linecount`.
 
-  Examples:
+  Example:
 
-  * For Kindle:
-
-    ```sh
-    fbink -t regular=/usr/java/lib/fonts/Caecilia_LT_65_Medium.ttf,bold=/usr/java/lib/fonts/Caecilia_LT_75_Bold.ttf,size=24,top=100,bottom=500,left=25,right=50,format "Hello **world**!"
+  * ```sh
+    fbink -t regular=$RFONT,bold=$BFONT,size=24,top=100,bottom=500,left=25,right=50,format "Hello **world**!"
     ```
 
-    Will use Caecilia to print "Hello world!" at 24pt in a display area starting from 100px down the top of the screen to 500px before the bottom of the screen, from 25px of the left edge of the screen until 50px before the right edge. Honoring the Markdown syntax, "Hello" will be printed with the Regular font style, while "world" will use the Bold font style. You will NOT be able to use obfuscated or encrypted fonts.
+    Will print "Hello world!" at 24pt in a display area starting from 100px down the top of the screen to 500px before the bottom of the screen, from 25px of the left edge of the screen until 50px before the right edge. Honoring the Markdown syntax, "Hello" will be printed with the Regular font style, while "world" will use the Bold font style. You will NOT be able to use obfuscated or encrypted fonts.
 
-  * For other than Kindle:
+    For Kindle try:
 
     ```sh
-    fbink -t regular=/mnt/onboard/fonts/NotoSans-Regular.ttf,bold=/mnt/onboard/fonts/NotoSans-Bold.ttf,size=24,top=100,bottom=500,left=25,right=50,format "Hello **world**!"
+    RFONT=/usr/java/lib/fonts/Caecilia_LT_65_Medium.ttf
+    BFONT=/usr/java/lib/fonts/Caecilia_LT_75_Bold.ttf
     ```
 
-    Will use NotoSans to print "Hello world!" at 24pt in a display area starting from 100px down the top of the screen to 500px before the bottom of the screen, from 25px of the left edge of the screen until 50px before the right edge. Honoring the Markdown syntax, "Hello" will be printed with the Regular font style, while "world" will use the Bold font style. You will NOT be able to use obfuscated or encrypted fonts.
+    And others (other than Kobo):
 
-  * Note for Kobo: this means you will NOT be able to use system fonts on Kobo, because they're all obfuscated.
+    ```sh
+    RFONT=/mnt/onboard/fonts/NotoSans-Regular.ttf
+    BFONT=/mnt/onboard/fonts/NotoSans-Bold.ttf
+    ```
+
+    Note for Kobo: this means you will NOT be able to use system fonts on Kobo, because they're all obfuscated.
 
 ### Options for printing an image (if compiled with `FBINK_WITH_IMAGE`)
 
--- a/API.md
+++ b/API.md
@@ -1,0 +1,385 @@
+# FBInk Command Line App Manual
+
+## Synopsis
+
+* ```sh
+  fbink [OPTIONS] [STRING ...]
+  ```
+
+  Print `STRING`s on your device's screen.
+
+* ```sh
+  fbink [-WHf] [-k] --refresh [top=NUM,left=NUM,width=NUM,height=NUM]
+  ```
+
+  Refresh the screen as per your specification, without touching the framebuffer.
+
+* ```sh
+  fbink [-WHf] --cls [top=NUM,left=NUM,width=NUM,height=NUM]
+  ```
+
+  Clear the screen (or a region of it).
+
+## Examples
+
+* ```sh
+  fbink -x 1 -y 10 "Hello World!"
+  ```
+
+  Prints 'Hello World!' on the eleventh line, starting at the second column from the left.
+
+* ```sh
+  fbink -pmh -y -5 "Hello World!"
+  ```
+
+  Prints 'Hello World!', highlighted (i.e., white on black with the default colors), centered & padded on both sides, on the fifth line starting from the bottom.
+
+* ```sh
+  fbink -pmM -y -8 "Hello World!"
+  ```
+
+  Prints 'Hello World!', centered & padded on both sides, eight lines above the center of the screen.
+
+## Options
+
+### Options affecting the message's position on screen
+
+* `-x NUM`, `--col NUM`
+
+  Begin printing `STRING` @ column `NUM` (Default: 0).
+
+  You might consider beginning at column 1 instead of 0, as column 0 (the leftmost one) may sometimes be slightly obscured by a bezel.
+
+  Use a negative value to count back from the right edge of the screen.
+
+* `-y NUM`, `--row NUM`
+
+  Begin printing `STRING` @ row `NUM` (Default: 0).
+
+  You might consider beginning at row 1 instead of 0, as row 0 (the topmost one) may sometimes be slightly obscured by a bezel, especially on Kobos.
+
+  Use a negative value to count back from the bottom of the screen.
+
+* `-X NUM`, `--hoffset NUM`
+
+  Adjust final text position by `NUM` pixels on the horizontal axis (Default: 0).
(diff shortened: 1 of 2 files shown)


API.md
```diff
@@ -0,0 +1,387 @@
...
+
+  * If compiled with `FBINK_FOR_KINDLE`:
+
+    ```sh
+    fbink -t regular=/usr/java/lib/fonts/Caecilia_LT_65_Medium.ttf,bold=/usr/java/lib/fonts/Caecilia_LT_75_Bold.ttf,size=24,top=100,bottom=500,left=25,right=50,format "Hello **world**!"
+    ```
+
+    Will use Caecilia to print "Hello world!" at 24pt in a display area starting from 100px down the top of the screen to 500px before the bottom of the screen, from 25px of the left edge of the screen until 50px before the right edge. Honoring the Markdown syntax, "Hello" will be printed with the Regular font style, while "world" will use the Bold font style. You will NOT be able to use obfuscated or encrypted fonts.
+
+  * If not compiled with `FBINK_FOR_KINDLE`:
```
NiLuJe, line 361: For humans, that actually means a Kobo build ;).

This is possibly applicable to other `FBINK_FOR` mentions in here: make it human readable an name the target platform explicitly instead of trying to grok the convoluted preprocessing rules for those ;p

API.md
```diff
@@ -0,0 +1,387 @@
...
+## Synopsis
+
+* ```sh
+  fbink [OPTIONS] [STRING ...]
+  ```
+
+  Print `STRING`s on your device's screen.
+
+* ```sh
+  fbink [-WHf] [-k] --refresh [top=NUM,left=NUM,width=NUM,height=NUM]
```
NiLuJe, line 12: A bit wary of mentioning `-k, --cls` here, especially since it's detailed below, and it makes the explanation below a lie, because it *will* modify the framebuffer ;).

It's also missing a whole bunch of supported flags.

API.md
```diff
@@ -0,0 +1,387 @@
...
+  ```
+
+  Refresh the screen as per your specification, without touching the framebuffer.
+
+* ```sh
+  fbink [-WHf] --cls [top=NUM,left=NUM,width=NUM,height=NUM]
```
NiLuJe, line 18: Ditto on the missing supported flags (not necessarily the same one, though. The listing in the printf wall should be accurate).

API.md
fulldecent: So, basically we have "for Linux" "for Kindle" and "for Kobo", which I can assume are mutually exclusive?

NiLuJe:
Not necessarily exhaustive comments after a (very) quick first look ;).

Other than specific nits, it looks good, and I'm in complete agreement with the reorganization mentioned in the "Differences" listing. (Besides the potential --refresh/--cls conflict commented upon above).

API.md
fulldecent: GitHub shows this weirdly. Are you referring to a definition in the Synopsis? Which one?

API.md
```diff
@@ -0,0 +1,388 @@
...
+
+* `-y`, `--row` `NUM`
+
+  Begin printing `STRING` @ row `NUM` (Default: 0).
+
+  You might consider beginning at row 1 instead of 0, as row 0 (the topmost one) may sometimes be slightly obscured by a bezel, especially on Kobos.
+
+  Use a negative value to count back from the bottom of the screen.
+
+* `-`, `--hoffset` `NUM`
```
NiLuJe, line 63: Missing short flag ;).

NiLuJe: `X`

API.md
```diff
@@ -0,0 +1,388 @@
...
+
+  Clear the full screen before printing.
+
+  Honors `-B`, `--background`; `-h`, `--invert`; `-H`, `--nightmode`; `-W`, `--waveform`; `-D`, `--dither`; `-b`, `--norefresh`; `-w`, `--wait`
+
+  Can be specified on its own, without any `STRING`.
+
+  NOTE: If your intent is to simply clear the screen and *nothing else*, use `-k`, `--cls` instead!
+
+* `-W`, `--waveform` `MODE` (not available if compiled with `FBINK_FOR_LINUX`)
```
NiLuJe, line 119: In the spirit of my earlier `FBINK_FOR` comment, that ones means "not an e-Ink device".

API.md
fulldecent: Could you please review this raw Markdown for a sec?

The description may not be lining up to the correct example line that you see. This is unambiguous when you read the code in the matrix/Markdown.

API.md
NiLuJe: Kindle/Kobo/Cervantes/reMarkable/PocketBook

And Linux for "not an e-Ink device" ;).

And, yeah, they're all mutually exclusive (the default being Kobo, not that it should matter, but it may help you follow the preprocessing ladders).

API.md
fulldecent: fixed at c7ef3759038a11145e85d1cda6e88b6598a931f5

API.md
fulldecent: The associated description is below the `---sh` block

API.md
fulldecent: The associated description is below the ---sh block

API.md
NiLuJe: I am actually reading the raw, haven't even looked at how it looks like processed ;).

API.md
NiLuJe: Yep:

https://github.com/NiLuJe/FBInk/blob/94305d2d2b40983dee12b2a236dac573ebbd556f/fbink_cmd.c#L117

NiLuJe: The "abort early" notice is somewhat relevant, because it means it can't be combined with other actions. (e.g., printing a string or an image).

(The `without touching the framebuffer` mention for `--refresh` essentially means the same thing, too).

API.md
NiLuJe: And, yeah, my bad, that one is *not* missing flags ;).

https://github.com/NiLuJe/FBInk/blob/94305d2d2b40983dee12b2a236dac573ebbd556f/fbink_cmd.c#L273

API.md
fulldecent: In the synopsis numbers 2 and 3, that usage is not advertising a string or other options, so the abort early is out of scope.

But in synopsis number 1, all of the options are available, even ones that collide, as so the "abort early" notice below applies, and it's in the Options specifications.

I think this is still right as-is. And-or I'm missing something.

NiLuJe: > In the synopsis numbers 2 and 3, that usage is not advertising a string or other options, so the abort early is out of scope.
> 

Technically correct, although I'm always in favor of making those thing explicitly clear ;).

> But in synopsis number 1, all of the options are available, even ones that collide, as so the "abort early" notice below applies, and it's in the Options specifications.
> 

Yep, wasn't talking about 1. at all ;).

> I think this is still right as-is. And-or I'm missing something.

It's certainly not *wrong* ;).

The only really missing stuff is the flags listing for 3..

API.md
fulldecent: Got it. Fixed in 7e4ab88586c01abc352eac313b266afb759e56e1

API.md
fulldecent: removed this "todo". 2b471773b12672fd198d9f31e2860eba8a39e7c2

API.md
fulldecent: /me screwed head back on. Okay, I think d9a0227f3ce0104d2fe9c171eca5cc916b546a26 is what we're looking for here...

API.md
fulldecent: Fixed in 7e4ab88586c01abc352eac313b266afb759e56e1

fulldecent:
First pass is complete.

I'm not currently aware any any outstanding TODOs in this code. (Although as follow on work (including help requested) three items in Work Plan could be done later.)

Therefore removing "Draft" and WIP from title.

API.md
```diff
@@ -356,23 +356,23 @@
...
+  * For Kindle:
 
     ```sh
     fbink -t regular=/usr/java/lib/fonts/Caecilia_LT_65_Medium.ttf,bold=/usr/java/lib/fonts/Caecilia_LT_75_Bold.ttf,size=24,top=100,bottom=500,left=25,right=50,format "Hello **world**!"
     ```
 
     Will use Caecilia to print "Hello world!" at 24pt in a display area starting from 100px down the top of the screen to 500px before the bottom of the screen, from 25px of the left edge of the screen until 50px before the right edge. Honoring the Markdown syntax, "Hello" will be printed with the Regular font style, while "world" will use the Bold font style. You will NOT be able to use obfuscated or encrypted fonts.
 
-  * If not compiled with `FBINK_FOR_KINDLE`:
+  * For other than Kindle:
```
NiLuJe, line 367: The only thing that differs here is the path being used in the example, it's not really tied to hardware/platform quirks, so I'd possibly amend those mentions to `on` Kindle/Kobo.

(Or any other wording you may find that conveys the fact that this is just an example using paths specific to *a* specific platform, namely Kindle and then Kobo).

API.md
```diff
@@ -0,0 +1,492 @@
...
+  Print `STRING`s on your device's screen.
+
+* ```sh
+  fbink [-fcWDHbhxyS] --image file=PATH,x=NUM,y=NUM,halign=ALIGN,valign=ALIGN,w=NUM,h=NUM,dither [--img PATH]
+  ```
+
+  Print image on your device's screen.
+
+* ```sh
+  fbink [-WHf] [-k] --refresh [top=NUM,left=NUM,width=NUM,height=NUM]
```
NiLuJe, line 18: That still leaves the `-k` mention here. If you specify -k, you're in fact calling --cls, which happens to *also* do a refresh (assuming you haven't disabled refreshes ^^), which isn't --refresh's use case (or codepath) ;).

API.md
fulldecent: How about 3e7ce40a7b1b1c8dcfa10a45af11defeabdf59ee?

API.md
fulldecent: How about? 7eb1ce102a382ffee4151bc12e83d5791aaaaddc

API.md
```diff
@@ -354,25 +354,29 @@
...
+    For Kindle try:
 
     ```sh
-    fbink -t regular=/mnt/onboard/fonts/NotoSans-Regular.ttf,bold=/mnt/onboard/fonts/NotoSans-Bold.ttf,size=24,top=100,bottom=500,left=25,right=50,format "Hello **world**!"
+    RFONT=/usr/java/lib/fonts/Caecilia_LT_65_Medium.ttf
+    BFONT=/usr/java/lib/fonts/Caecilia_LT_75_Bold.ttf
     ```
 
-    Will use NotoSans to print "Hello world!" at 24pt in a display area starting from 100px down the top of the screen to 500px before the bottom of the screen, from 25px of the left edge of the screen until 50px before the right edge. Honoring the Markdown syntax, "Hello" will be printed with the Regular font style, while "world" will use the Bold font style. You will NOT be able to use obfuscated or encrypted fonts.
+    And others (other than Kobo):
```
NiLuJe, line 372: That's the spirit ;).

Although, this *is* actually perfectly viable on Kobo. The warning below applies to *system* fonts, the example specifically uses the path to *user* fonts, to avoid that ;).

API.md
NiLuJe: Nope, just getting rid of the `[-k]` mention would do it ;). The description was perfectly accurate.

fulldecent: Fixed in 2871edc3d4b65ec2d8cbe8f62fd241efff4b05ae

fulldecent:
I think all outstanding comments here are addressed as per the API.md file. Good to merge?

NiLuJe:
Got a bit sidetracked by a KOReader bug hunt, so, probably in a day or two, yeah ;).

NiLuJe:
Well, that was an even busier week than expected, sorry ;p.

Looks good, thanks!

I'll deal with the remaining inconsistencies w/ the printf output ;).

PR #64, opened 2021-07-11
Kobo Elipsa support (a.k.a, sunxi)

NiLuJe:
Well, that was *not* fun.

---
This change is [Reviewable](https://reviewable.io/reviews/niluje/fbink/64)


NiLuJe:
A few significant quirks:

* The framebuffer itself is now almost entirely meaningless. We basically only need it to tell us the screen's dimensions. This makes the whole FBFD_AUTO workflow kind of tricky, because we need a whole other set of things besides a fd fd to make it work... All of that extra stuff is kept internal, and may or may not behave entirely properly when using FBFD_AUTO. (But i'll make sure it does when I update KFMon).

* This also means that dealing with rotation has become even more hellish than usual, since you don't actually have any way of telling what's the layout of what is currently on the screen. Instead, I had to resort to some dirty hackery to poke directly at the accelerometer to tell us how the device is *physically* laid out. But that doesn't necessarily match the screen's layout... And, as a cherry on the cake, if you do attempt to display stuff in a different layout, all hell breaks loose, and none of the available workarounds are all that great. (Follow the breadcrumbs around `fbink_toggle_sunxi_ntx_pen_mode` in the header for more details).

* I'm probably forgetting a few things, because I've lost count of how many times doing stuff wrong caused a kernel OOPS/BUG/hang/softlock/reboot. `-_-"`

NiLuJe:
And now, ping @gtalusan, because I come bearing bug reports ;p.

In no particular order:

Kernel issues:

* The `ION_HEAP_QUERY` is subtly broken, because of a mainline issue. c.f., https://github.com/NiLuJe/FBInk/blob/24acf62e8eb495775051dd56ab041b5a8400718d/utils/ion_heaps.c#L111-L116
(Granted, we don't need it, and, as the code above shows, we can get it to work, but it came in handy to triple-check a few ION things, and the fix is trivial ;)).

* There's a rather nasty issue around the `EINK_DITHERING_` constants: they all share a bit with `EINK_DITHERING_Y1`, which happens the bit that the if ladder checks for *first* in `eink_image_process_thread` @ `drivers/video/fbdev/sunxi/disp2/disp/de/eink_buffer_manager.c`, which effectively means that every dithering flag actually ends up being detected as this one.
c.f., https://github.com/NiLuJe/FBInk/blob/24acf62e8eb495775051dd56ab041b5a8400718d/fbink.c#L2619-L2624

* Gnarlier, there's a similar bitmask issue with the *waveform* constants, this time: all the constants *strictly greater than* `EINK_GLD16_MODE` share at least one bit with other modes. In particular, the GLK/GCK ones share bits with A2, and, at a quick glance, there *is* code that checks and takes A2-specific code-paths *after* night-specific codepaths...

The issue is slightly gnarlier than the dithering one in that there *is* enough space in the first 16 bits (even without moving `RECT` & `AUTO` to higher bits) to fit all of these, but a *lot* of the code masks the wrong amount of bytes manually (`& 0xFF`) instead of using the `GET_UPDATE_MODE` macro...

* In the same vein, and possibly related, `GC4L` & `CLEAR` currently crash the driver (first causes a timeout and then silent failures, and the device becomes *extremely* reluctant to soft reboot).

* Something slightly more trivial than I just noticed when cruising by: in `ring_buffer_manager_init` (@ `drivers/video/fbdev/sunxi/disp2/disp/de/eink_buffer_manager.c`), in the error codepath, a buffer is freed with a wrong size (`*`) 4 @ L3493 while it's allocated without a multiplier @ L3340...

----

nickel/pickel quirks:

* pickel appears to be setting the `heap_id_mask` field in the `ION_IOC_ALLOC` ioctl to `ION_HEAP_MASK_CARVEOUT|ION_HEAP_MASK_DMA`. Is that on purpose (there is enough space left in the carveout heap)? Right now DMA gets picked simply because it has higher priority thanks to its higher id.
I haven't straced nickel yet, but it does also end up with an alloc in the DMA heap.

* pickel passes a pointer instead of a value for the `flags` arg in the `ION_IOC_ALLOC` ioctl. Again, haven't checked nickel.

* Both nickel & pickel pass a pointer instead of a value for the `cfa_use` arg in the `DISP_EINK_UPDATE2` ioctl.

The ion stuff was sniffed via a patched strace, as was the disp stuff, although it's also evident in debug logging if you raise disp's debug level.

gtalusan:
Nickel uses its own accelerated dithering instead of the kernel's.  You should too.

Don't confuse pointers with uninitialized/doesn't-crash-then-don't-care values.

gtalusan:
Also CARVEOUT | DMA is probably definitely a bug on our side.  CARVEOUT is the way to go.  💯 on that.

NiLuJe:
> Nickel uses its own accelerated dithering instead of the kernel's. You should too.

I usually am, but the EPDCv2 available on Mk. 7 was surprisingly decently correct ;).

> 
> Don't confuse pointers with uninitialized/doesn't-crash-then-don't-care values.

I was on the fence, but the fact that they appeared to be close to actual useful pointers *and* deref to "correct" (usually 0 ;p) values was strange ^^.

NiLuJe:
> Also CARVEOUT | DMA is probably definitely a bug on our side. CARVEOUT is the way to go. 100 on that.

Switched to the carveout in 934616f, at a very quick glance nothing new horribly blew up, thanks ;).

```
┌─(ROOT@europa:pts/1)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~)─┐
└─(2.04:19%:21:55:100%:#)── cat /sys/kernel/debug/ion/heaps/carveout                                                                                                                                                                                                                                                                                   ──(Sun, Jul 11)─┘
          client              pid             size
----------------------------------------------------
        disp_ion                1         97714176
    finger_trace             2908          2629632
----------------------------------------------------
orphaned allocations (info is from last known client):
----------------------------------------------------
  total orphaned                0
          total         100343808
   deferred free                0
----------------------------------------------------
```

PR #65, opened 2021-08-12
Add more functionality to the API

Rain92:
These are my proposed changes for the FBInk API.

Some new members to the FBInkState struct:
~~uint8_t current_rota_native;                 // native screen rotaiton; vInfo.rotate (current rotation, c.f., <linux/fb.h>)
uint8_t current_rota_canonical;            // canonical screen rotation~~
 uint32_t    screen_stride;                     // screen line length in bytes;

A new method to access the underlying framebuffer:
// Grants direct access to the frame buffer pointer as well as the size of the frame buffer allocation.
// If the framebuffer is not yet allocated it will do so and return the result of the operation.
FBINK_API int
    fbink_get_fb_pointer(int fbfd, FBPtrInfo *fbInfo);
    

A new method to set the framebuffer rotation and bpp.
The code is mostly copied from fbdepth and probably needs some tinkering:
// Sets the framebuffer bits per pixel and rotation and invoke a reinit afterwards
// rota will be the rotation in natve format; use fbink_rota_canonical_to_native to convert it to canonical
// bpp will remain unchanged if the value is < 8
// req_gray will remain unchanged if the value is < 0
// on sunxi devices it will invoke fbink_sunxi_ntx_enforce_rota
FBINK_API int
    fbink_set_fb_info(int fbFd, int32_t bpp, int8_t rota, int32_t req_gray,
               const FBInkConfig* restrict fbink_cfg);

---
This change is [Reviewable](https://reviewable.io/reviews/niluje/fbink/65)

### DIFF
--- a/fbink.c
+++ b/fbink.c
@@ -10847,6 +10847,199 @@ int
 }
 
 
+#ifdef FBINK_FOR_KINDLE
+static orientation_t
+    linuxfb_rotate_to_einkfb_orientation(uint32_t rotate)
+{
+    switch (rotate) {
+        case FB_ROTATE_UR:
+            return orientation_portrait;
+        case FB_ROTATE_CW:
+            return orientation_landscape;
+        case FB_ROTATE_UD:
+            return orientation_portrait_upside_down;
+        case FB_ROTATE_CCW:
+            return orientation_landscape_upside_down;
+        default:
+            // Should never happen.
+            return orientation_portrait;
+    }
+}
+
+static uint32_t
+    einkfb_orientation_to_linuxfb_rotate(orientation_t orientation)
+{
+    switch (orientation) {
+        case orientation_portrait:
+            return FB_ROTATE_UR;
+        case orientation_portrait_upside_down:
+            return FB_ROTATE_UD;
+        case orientation_landscape:
+            return FB_ROTATE_CW;
+        case orientation_landscape_upside_down:
+            return FB_ROTATE_CCW;
+        default:
+            // Should never happen.
+            return FB_ROTATE_UR;
+    }
+}
+#endif
+
+int
+    fbink_set_fb_info(int fbfd, int32_t bpp, int8_t rota, int32_t req_gray,
+               const FBInkConfig* restrict fbink_cfg)
+{
+    // Set variable fb info
+    // Bitdepth; leave unchanged if < 8
+    if (bpp >= 8) {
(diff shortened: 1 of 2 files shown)


NiLuJe:
I'll have more time this week-end, so, very, very quick comments:

* Not a fan of the split `current_rota_native`/`current_rota_canonical`, as there's nothing private about one of these: it's just the other run through a public API call ;).

* I'd probably mark the exported pointer `const` to document the fact that the pointer itself should be kept around (or even export a pointer to internal fbPtr to be able to track the mapping state; with the same const markings).

* Will review the fbdepth stuff, that's definitely long overdue ;). (utils/dump.c has similar code that was a bit clunkier because not relying on private APIs ;p).

* The get_fb_data stuff probably ought to export const pointers to const data for vInfo & fInfo, because PocketBook suffers from the same inconsistencies we discussed ;).

NiLuJe:
>     * The get_fb_data stuff probably ought to export const pointers to const data to vInfo & fInfo, because PocketBook suffers from the same inconsistencies we discussed ;).

Then again, the rota/bitdepth stuff isn't really supported on PB (as it's even quirkier than on Kobo), and the only thing one could really have missed was indeed `line_len` & `smem_len` ;).

Plus, I really like the `get_fb_ptr` name, and that wouldn't fit as well if it did other stuff ;p.

Rain92:
I agree it's not really nesassary to expose vInfo & fInfo right now (except for the color format in vinfo but let's just ignore that).
I can rename it to get_fb_ptr. 
Do you propose to put the allocation size into the FBInkState struct instead?

Rain92:
I am not sure how to mark the exported pointer as const tho as it has to be written somewhere if it is passed as an argument.
And if it is passed as the return value it would be a bit wanky because the method has to return an exit code.

NiLuJe:
I was vaguely thinking of something along `char * const get_fb_pointer(int fbfd, sizer_t* alloc_size)`, i.e., the alloc size as an outptr, and returning `NULL` on error ;).

NiLuJe:
As far as set_fbinfo is concerned (don't mind the names I'm using here, nothing's set in stone ;p), I'd move it to fbink_rota_quirks.c.

As for its actual code, I've got a few notes jotted down, but I'll probably simply handle some minor cleanup post-merge, as the API looks fine ;).

NiLuJe:
Thanks!

PR #69, opened 2022-06-23
[PW5 MTK] Fix swipe animations with a tiny refresh area

yparitcher:
koreader/koreader#9245 koreader/koreader-base/pull/1494

The simplest way to trigger this is `fbink -K direction=LEFT,steps=12 -s width=11`

The MTK Driver will crash if given a area smaller than the number of steps. If direction is L/R and w is smaller or if it is U/D and h is smaller Being as one generally will only want animations on a larger area, and I am too lazy to test for direction, disable animations when w or h is less than steps.

I did not add any error handling etc, feel free to spice it up.

---
This change is [Reviewable](https://reviewable.io/reviews/niluje/fbink/69)

### DIFF
--- a/fbink.c
+++ b/fbink.c
@@ -2142,12 +2142,18 @@ static int
 
 	// If requested, enable the refresh animation, as setup by fbink_mtk_set_swipe_data.
 	if (fbink_cfg->is_animated) {
-		update.swipe_data    = mtkSwipeData;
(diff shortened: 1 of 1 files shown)


yparitcher:
Not tested

NiLuJe:
Oh, makes sense, thanks for the detective work ;).

NiLuJe:
Thanks!

PR #77, opened 2024-02-13
README: Link to a couple of Rust bindings

Szybet:
This change is [Reviewable](https://reviewable.io/reviews/NiLuJe/FBInk/77)

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,9 @@ If you don't actually need a PNG file and just want to play with in-memory fb du
 
 So that everyone gets to have fun, even if you can't stand C!
 
-[Rust](https://www.rust-lang.org/): [fbink-sys](https://github.com/Szybet/fbink-sys) by [@Szybet](https://github.com/Szybet)
+[Rust](https://www.rust-lang.org/): 
+- [fbink-sys](https://github.com/Szybet/fbink-sys) by [@Szybet](https://github.com/Szybet) (Bare bindgen bindings)
+- [fbink-rs](https://github.com/sublipri/fbink-rs) by [@sublipri](https://github.com/sublipri) (Bare bindgen bindings, some safe wrappers, dependent on [cross](https://github.com/cross-rs/cross))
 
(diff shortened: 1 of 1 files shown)


sublipri:
I've been tinkering with some Rust bindings too over the last few weeks. I couldn't get it working with musl until seeing your use of the inkbox toolchain, so thanks for that! Took a different approach with bindgen and called clang from `build.rs` instead of using the makefile. I had a basic example working using the `arm-unknown-linux-gnueabihf` target, but ended up with problems due to the outdated kernel/libc. 

Ended up using a toolchain from the musl website to make it easier to download during the build process. I'm not sure if it's identical to the inkbox one - same version as mentioned in the [commit](https://github.com/Kobo-InkBox/kernel/commit/8f76c621d15d5a6ae3322d10ed39bd5b57791d4f#diff-d367b4820bce6e9455010690b0662aa5852441428c7f244d62cf5b0898ce39e9), but the unpacked filesize wasn't identical. I have very limited experience with C, so the way I've done things might not be ideal, but it seems to be working on a Glo and Sage in limited testing.

Made some progress with a safe Rust wrapper, but it's very incomplete and I'm not sure if I'll manage to work on it much further aside from anything I need to add for my use case.

https://github.com/sublipri/fbink-rs

Szybet:
That's very cool
Cross has it's problems, so I would prefer to avoid it

Szybet:
Can I add your repo to this pull request as an alternative to mine?

sublipri:
Sure! That'd be cool. I'll add a link to your repo in my readme too as an alternative if people have issues with cross.

NiLuJe:
@Szybet I'm finally back, feel free to update this with both links as you mentioned above ;).

Szybet:
Done, I don't really know how you want to format that so I went with a list

I also added link to your repo in my repo too :D

NiLuJe:
> so I went with a list

Sounds good to me, thanks ;).

PR #82, opened 2025-01-17
[feat] Add support for loading OpenType fonts from memory.

notmarek:
This change is [Reviewable](https://reviewable.io/reviews/NiLuJe/FBInk/82)

### DIFF
--- a/fbink.c
+++ b/fbink.c
@@ -5809,7 +5809,7 @@ int
 #endif    // FBINK_WITH_OPENTYPE
 }
 
-// Load font fron from memory. Up to four font styles may be used per FBInkOTConfig instance.
+// Load font from memory. Up to four font styles may be used per FBInkOTConfig instance.
 int fbink_add_ot_font_from_memory_v2(const unsigned char* data UNUSED_BY_MINIMAL, FONT_STYLE_T style UNUSED_BY_MINIMAL, FBInkOTConfig *restrict cfg UNUSED_BY_MINIMAL) {
 #ifdef FBINK_WITH_OPENTYPE
     // Start by allocating an FBInkOTFonts struct, if need be...
--- a/fbink.c
+++ b/fbink.c
@@ -5626,8 +5626,16 @@ static __attribute__((cold)) const char*
 
 // Load OT fonts fot add_ot_font
 static __attribute__((cold)) int
-    add_ot_font_data(const unsigned char* data, FONT_STYLE_T style, FBInkOTFonts* restrict ot_fonts)
+    	add_ot_font_data(const unsigned char* data, FONT_STYLE_T style, FBInkOTFonts* restrict ot_fonts)
 {
+
+	// Init libunibreak the first time we're called
+	if (!otInit) {
+		init_linebreak();
+		LOG("Initialized libunibreak");
+	}
+	otInit = true;
+
     stbtt_fontinfo* font_info = calloc(1U, sizeof(stbtt_fontinfo));
 	if (!font_info) {
 		PFWARN("Error allocating stbtt_fontinfo struct: %m");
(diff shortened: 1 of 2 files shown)


fbink.c
```diff
@@ -5624,19 +5624,10 @@ static __attribute__((cold)) const char*
 	}
 }
 
-// Load OT fonts for fbink_add_ot_font & fbink_add_ot_font_v2
+// Load OT fonts fot add_ot_font
```
NiLuJe, line 5627:
```suggestion
// Load OT fonts for add_ot_font
```

fbink.c
```diff
@@ -5764,6 +5771,20 @@ int
 #endif    // FBINK_WITH_OPENTYPE
 }
 
+// Load fron from memory. Up to four font styles may be used by FBInk at any given time.
```
NiLuJe, line 5774:
```suggestion
// Load font from memory. Up to four font styles may be used by FBInk at any given time.
```

fbink.c
```diff
@@ -5788,6 +5809,26 @@ int
 #endif    // FBINK_WITH_OPENTYPE
 }
 
+// Load font fron from memory. Up to four font styles may be used per FBInkOTConfig instance.
```
NiLuJe, line 5812:
```suggestion
// Load font from memory. Up to four font styles may be used per FBInkOTConfig instance.
```

fbink.c
```diff
@@ -5741,13 +5696,65 @@ static __attribute__((cold)) int
...
+			free(data);
+			fclose(f);
+			otInit = false;
+			WARN("Error reading font file `%s`", filename);
+			return ERRCODE(EXIT_FAILURE);
+		}
+		fclose(f);
+	}
+	int result = add_ot_font_data(data, style, ot_fonts);
+	free(data);
```
NiLuJe, line 5755: That's a problem ;).

(The data is still live, stbtt doesn't make a copy, it just keeps a reference to that same pointer).

fbink.c
```diff
@@ -5645,70 +5636,34 @@ static __attribute__((cold)) int
...
 	} else if (fontcount > 1) {
-		LOG("Font file `%s` appears to be a font collection containing %d fonts, but we'll only use the first one!",
-		    filename,
+		LOG("Font file appears to be a font collection containing %d fonts, but we'll only use the first one!",
 		    fontcount);
 	}
 	// Then, get the offset to the first font
 	const int fontoffset = stbtt_GetFontOffsetForIndex(data, 0);
 	if (fontoffset == -1) {
-		free(data);
```
NiLuJe, line 5701: In the same vein, this needs to be `free(font_info->data);`

Which impliers that we take ownership of the pointer, which might be a problem for your usecase (e.g., pointer to static content), so this may need to be more complex to add a way to say "nope, this data is immutable, don't ever try to free it).

fbink.c
```diff
@@ -5645,70 +5636,34 @@ static __attribute__((cold)) int
...
+    stbtt_fontinfo* font_info = calloc(1U, sizeof(stbtt_fontinfo));
 	if (!font_info) {
 		PFWARN("Error allocating stbtt_fontinfo struct: %m");
-		free(data);
 		return ERRCODE(EXIT_FAILURE);
 	}
 	// First, check if we can actually find a recognizable font format in the data...
 	const int fontcount = stbtt_GetNumberOfFonts(data);
 	if (fontcount == 0) {
-		free(data);
```
NiLuJe, line 5689: ditto

fbink.c
```diff
@@ -5645,70 +5636,34 @@ static __attribute__((cold)) int
...
-			WARN("Error reading font file `%s`", filename);
-			return ERRCODE(EXIT_FAILURE);
-		}
-		fclose(f);
-	}
-	stbtt_fontinfo* font_info = calloc(1U, sizeof(stbtt_fontinfo));
+    stbtt_fontinfo* font_info = calloc(1U, sizeof(stbtt_fontinfo));
 	if (!font_info) {
 		PFWARN("Error allocating stbtt_fontinfo struct: %m");
-		free(data);
```
NiLuJe, line 5683: ditto

fbink.c
NiLuJe: Weeelll, except `font_info->data` only exists *after* `stbtt_InitFont` ;p.

The point about ownership still stands, though ;p.

PR #83, opened 2025-05-06
allow linking the final binary with `-Wl,--gc-sections`

benoit-pierre:
Which means making sure partial linking is actually done without this option.

- - -
This change is [Reviewable](https://reviewable.io/reviews/NiLuJe/FBInk/83)


NiLuJe:
Thanks!

NiLuJe:
I'll bump KoboUSBMS when I get back tonight ;)

PR #87, opened 2026-04-07
update cdecls for latest ffi-cdecl version

benoit-pierre:
Cf. koreader/ffi-cdecl#19.

- - -
This change is [Reviewable](https://reviewable.io/reviews/NiLuJe/FBInk/87)


NiLuJe:
Nice! Thanks!

PR #91, opened 2026-08-05
ci: fix koxtoolchain install

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/NiLuJe/FBInk/91)


benoit-pierre:
```
tar: Archive is compressed. Use --zstd option
tar: Error is not recoverable: exiting now
```
Thanks tar, that's very useful!

